PR #75, opened 2014-08-18
Partition [wip]

tmcw:
Right now this falls into a fairly serious problem - if you have an
array of length=4 and want 3 chunks, it's not possible to get 3 chunks
with the `chunk()` underlying function, since the input must be integer,
1 will give 4 and 2 will give 2.

So, we either need to change the expectation for the function, saying
that it will _aim_ for the desired number of partitions, or not use
`chunk()` under this.

### DIFF
--- a/src/simple_statistics.js
+++ b/src/simple_statistics.js
@@ -666,6 +666,23 @@
         return output;
     }
 
+    // # partition
+    //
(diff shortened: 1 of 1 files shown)


tmcw:
Closing: needs a better defined problem.

PR #84, opened 2015-02-07
Fix rounding error in cumulative_std_normal_probability lookup.

ericfischer:
And add a numerical approximation to error_function() instead of only having a table lookup.


tmcw:
Awesome! Can you add a test for the error function too so that code coverage is still 100%?

ericfischer:
Sure, here's the new test (and a correction for my documentation error).

tmcw:
Thanks!

PR #99, opened 2015-05-31
Use larger umd declaration to be valid even in the case of use strict

### LINKED ISSUE #93: Working on Live Demo but not working from my machine on Chrome(Chrome Version  42.0.2311.90 m)

tmcw:
/cc @NerdYouDontKnow @stevemandl  mind trying out this branch and confirming that it'll fix #93

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,6 @@
 {
   "env": {
+    "amd": true,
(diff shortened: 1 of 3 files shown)


NerdYouDontKnow:
confirmed

PR #103, opened 2015-06-09
implement a simple perceptron classifier

ajschumacher:
This adds `.perceptron()` for making a single-layer perceptron classifier with `.train()` and `.predict()` methods. Tests are in `perceptron.test.js`.

I haven't added anything to the README etc., but if this looks good that could also be done.

### DIFF
--- a/src/simple_statistics.js
+++ b/src/simple_statistics.js
@@ -276,7 +276,7 @@
         perceptron_model.predict = function(features) {
             // Only predict if previously trained
             // on the same size feature array(s).
-            if (features.length != w.length) return null;
+            if (features.length !== w.length) return null;
             // Calculate the sum of features times weights,
             // with the bias added (implicitly times one).
             var score = 0;
@@ -285,7 +285,7 @@
             }
             score += b;
             // Classify as 1 if the score is over 0, otherwise 0.
-            return 0 < score ? 1 : 0;
+            return score > 0 ? 1 : 0;
         };
 
         // ## Train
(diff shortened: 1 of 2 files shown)


src/simple_statistics.js
```diff
@@ -254,6 +254,83 @@
...
+    // # [Perceptron Classifier](http://en.wikipedia.org/wiki/Perceptron)
+    //
+    // This is a single-layer perceptron classifier that takes
+    // arrays of numbers and predicts whether they should be classified
+    // as either 0 or 1 (negative or positive examples).
+    function perceptron() {
+        var perceptron_model = {},
+            // The weights, or coefficients of the model;
+            // weights are only populated when training with data.
+            w = [],
```
tmcw: could this be called `weights` instead of `w`?

src/simple_statistics.js
```diff
@@ -254,6 +254,83 @@
...
+                for (var i = 0; i < w.length; i++) {
+                    w[i] += gradient * features[i];
+                }
+                b += gradient;
+            }
+            return perceptron_model;
+        };
+
+        // Conveniently access the weight array.
+        perceptron_model.w = function() {
```
tmcw: similarly, should this be `weights`?

src/simple_statistics.js
ajschumacher: Sure could! I'll also change `b` to `bias`.

ajschumacher:
@tmcw Updated! :)

tmcw:
Awesome, merging!

PR #104, opened 2015-07-03
update/sync docs/add perceptron to README

ajschumacher:
This syncs the automatically generated documentation with the current version of source (after minor tweaks for readability, etc.) and adds `perceptron` documentation to README.md (and API.md).

Also I noticed that the `make test` was talking about mocha, which didn't seem to make sense, so I took that out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 components
 build
 node_modules
+docs/public/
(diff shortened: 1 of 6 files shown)


tmcw:
:+1:

PR #108, opened 2015-07-09
Calculate standard normal table

### LINKED ISSUE #107: Calculate Standard Normal Table
@mourner points out that we might not want to hardcode a standard normal table as we do now, but calculate it. I'm open to this as long as the dynamically-generated version is exactly equal to our table currently.

mourner:
Closes #107. The tables are identical. The calculation itself is sub-1ms.

We could improve the current precision further by either bumping the table size, the number of digits and iterations. 

Or, we could expose the actual function with memoization. A non-rounding calculating function call takes about 3-4 times more than the table call currently with 20 iterations.


src/standard_normal_table.js
```diff
@@ -8,74 +8,30 @@
...
+    for (var i = 1; i < 15; i++) {
+        tmp *= z * z / (2 * i + 1);
+        sum += tmp;
+    }
+    return Math.round((0.5 + (sum / SQRT_2PI) * Math.exp(-z * z / 2)) * 1e4) / 1e4;
+}
+
+var standardNormalTable = [];
+
+for (var z = 0; z <= 3.09; z += 0.01) {
```
tmcw: where does 3.09 come from?

mourner: It's the max the original table has. :) We can change to 3 and update the tests.

tmcw: yeah, the question is then why does the original table stop then...

mourner: Probably because it's so close to `1` for values after `3` that the error becomes acceptable. See the graph here https://en.wikipedia.org/wiki/File:Normal_Distribution_CDF.svg, red line.

tmcw:
:+1: awesome

PR #115, opened 2015-07-26
Implement equalIntervalBreaks. Fixes #57

### LINKED ISSUE #57: equal interval classification
Like jenks, but splits on equal intervals. Might be too trivial, but I have been finding myself building this out all the time when using ss.

tmcw:
cc @morganherlocker could you do a quick code review?


morganherlocker:
:+1: looks great

Issue #124, opened 2015-08-13
Proof Ckmeans


tmcw:
* Get a instrumented version of the R library working
* Add this result as a testcase https://github.com/simple-statistics/simple-statistics/pull/118#issuecomment-130608728 and figure out what's going on with it

llimllib:
Here's a much smaller test case:

```
$ tail -n3 test/ckmeans.test.js
    t.deepEqual(cK([0, 3, 4], 2), [[0], [3,4]]);
    t.end();
});
10:16 PM nikkisix.local:~/code/simple-statistics  master
$ node test/ckmeans.test.js
TAP version 13
# C k-means
ok 1 exports fn
not ok 2 should be equivalent
  ---
    operator: deepEqual
    expected: [ [ 0 ], [ 3, 4 ] ]
    actual:   [ [ 0, 3 ], [ 4 ] ]
    at: Test.<anonymous> (/Users/llimllib/code/simple-statistics/test/ckmeans.test.js:34:7)
  ...
```

It's clear that [(0), (3,4)] has a much smaller cluster distance than [(0,3), (4)], so *something* is wrong with our algorithm.

To find that test case, I wrote [some property-based tests](https://gist.github.com/llimllib/63fcba0492a3ac4a27bf) with @DRMaciver's excellent hypothesis library. Although the testing code needs work[1], this seems to be a good counterexample that fails on both our algorithms in the same way.

[1]: right now if you run it and it finds a counterexample, the next time you run it it may yell at you about flaky tests, because of how I'm generating the data. Working on fixing that right now.

llimllib:
OK, I improved example generation and updated the gist. Another small example is: `ckmeans([-1, 0, 0], 2) = [[-1, 0], [0]]`, which should be [[-1], [0,0]].

llimllib:
I also just noticed that the tests in the test_ckmeans.js file are indeed wrong, as the comment in there suggests.

`ckmeans([1,2,2,3], 3)` should be `[[1], [2, 2], [3]]`, for example

llimllib:
I think we're failing to consider the first element of the array, because this code: 

```
        for (var sortedIdx = Math.max(cluster, 1);
             sortedIdx < sorted.length;
             sortedIdx++) {
```

never considers sorted_idx 0. Does that seem plausible to you?

llimllib:
(edit: no, not plausible)

DRMacIver:
[...]

PR #125, opened 2015-08-16
Fix ckmeans

llimllib:
* the `withinss` was being calculated incorrectly; `((data_idx - 1) / data_idx) * squared_difference` ignores the first data point (`0/1`) instead of multiplying it by `1/2`
* the mean was being calculated incorrectly. On the second data point, where `data_idx` is 1, `first_cluster_mean` was being set to the sum of the first two numbers instead of half that sum, and so on for each index. subtracting one from `data_idx` does the trick.

Related to, but does not complete, #124 .


tmcw:
:raised_hands: awesome!

PR #130, opened 2016-01-29
add equalIntervalBreaks to index.js

rsbowman:
Hi, I noticed that equalIntervalBreaks isn't in the ss namespace.  Here's a patch that adds it.  Thanks for the cool software!


tmcw:
:+1: great, thanks!

PR #141, opened 2016-03-14
Implement ss.product()

JamesMcGuigan:
Implemented ss.product() as the multiplication of all elements in an array. Based on ss.sum()


tmcw:
:+1: awesome, thanks!

PR #146, opened 2016-03-24
Use quickselect to significantly speed up median and quantile

### LINKED ISSUE #145: Improve median and quantile performance by using a selection algorithm
Median and quantile calculation performance can be improved from `O(N log N)` to `O(N)` (a significant speedup) by using a [selection algorithm](https://en.wikipedia.org/wiki/Selection_algorithm) ("k-th element"). There's a fast implementation [here](https://github.com/mourner/quickselect).

cc @tmcw

mourner:
A work in progress, closes #145. The first commit implements median using quickselect, making it much faster — from `O(N log N)` to `O(N)` for unsorted input.

For a sample benchmark on 1 million random numbers, the current implementation runs in 755ms. The new one is 28ms — 27 times faster.

Remaining tasks if the approach is approved:

- [x] rewrite quantile using quickselection
- [x] pull quickselect into the repo
- [ ] clean up new quantile implementation, add comments
- [ ] add `nthElement` to the API (a generalization of `min` and `max` for any `n`)

cc @tmcw HAPPY BIRTHDAY!!! :cake: :cake: :cake:

### DIFF
--- a/src/median.js
+++ b/src/median.js
@@ -1,8 +1,7 @@
 'use strict';
 /* @flow */
 
-var medianSorted = require('./median_sorted'),
-    numericSort = require('./numeric_sort');
+var quickselect = require('quickselect');
 
 /**
  * The [median](http://en.wikipedia.org/wiki/Median) is
@@ -22,9 +21,28 @@ var medianSorted = require('./median_sorted'),
  * median(incomes); //= 3.5
  */
 function median(x /*: Array<number> */)/*:number*/ {
-    // Sorting the array makes it easy to find the center, but
-    // use `.slice()` to ensure the original array `x` is not modified
-    return medianSorted(numericSort(x));
+    // The median of an empty list is NaN
+    if (x.length === 0) { return NaN; }
+
+    var copy = x.slice();
+    var k;
+
+    // If the length of the list is odd, it's the central number
+    if (x.length % 2 === 1) {
+        k = (x.length - 1) / 2;
+        // Rearrange so that k-th element is k-th smallest
+        quickselect(copy, k);
+        return copy[k];
+
+    // Otherwise, the median is the average of the two numbers at the center of the list
+    } else {
+        k = x.length / 2 - 1;
+        // Rearrange so that k-th element is k-th smallest
+        quickselect(copy, k);
+        // Rearrange the remaining half so that the first element is smallest
+        quickselect(copy, k + 1, k + 1);
+        return (copy[k] + copy[k + 1]) / 2;
+    }
 }
 
 module.exports = median;
--- a/src/median.js
+++ b/src/median.js
@@ -1,7 +1,7 @@
 'use strict';
 /* @flow */
 
-var quickselect = require('quickselect');
+var quantile = require('./quantile');
 
 /**
  * The [median](http://en.wikipedia.org/wiki/Median) is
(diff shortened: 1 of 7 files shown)


mourner:
The new quantile implementation is much more sophisticated — it uses a divide and conquer quickselect for the quantile array case, and needs some clean up in implementation to dry up the code. 

But it's also much faster, improving from `O(N log N)` to  `O(N)` for the single case and to `O((K + N) log K)` for the array case, where K is the number of requested quantiles.

Benchmark results for 1 million points:

- 170ms for 1000 quantiles
- 110ms for 100 quantiles
- 70ms for 10 quantiles

Compared to ~750ms for all three cases with the sort-based implementation.

mourner:
BTW, just realized than median can be implemented simply as `quantile(arr, 0.5)`.

tmcw:
Awesome, thanks! What would you think of pulling (a version of) the quickselect algorithm into simple-statistics? Right now we don't have any dependencies (only devDependencies) and have the guarantee that all code is written in roughly the same style and the same level of commenting. I'd love to keep simple-statistics that way - avoiding runtime dependencies for simplicity and keeping all code documented with no black boxes.

mourner:
@tmcw sure! I extracted `quickselect` just recently, it's still a part of `rbush` code. We can pull it here too. 

Some of the code added requires a pretty verbose explanation — I'll take a pass at it, but it'll probably require some edits from you to make it clear. 

Any comments/questions on the code?

src/median.js
```diff
@@ -22,9 +21,7 @@ var medianSorted = require('./median_sorted'),
  * median(incomes); //= 3.5
  */
 function median(x /*: Array<number> */)/*:number*/ {
-    // Sorting the array makes it easy to find the center, but
-    // use `.slice()` to ensure the original array `x` is not modified
-    return medianSorted(numericSort(x));
+    return +quantile(x, 0.5);
```
tmcw: Was flow unable to infer the type of quantile's return value without the `+`?

mourner: Yes, because `quantile` returns either a number or an array, but we're only interested in the `number` case here.

src/quantile.js
```diff
@@ -29,20 +29,81 @@ var numericSort = require('./numeric_sort');
...
+        return Math.ceil(idx) - 1;
+    } else if (len % 2 === 0) {
+        // If the list has even-length, we'll return the middle of two indices
+        // around quantile to indicate that we need an average value of the two
+        return idx - 0.5;
+    } else {
+        // Finally, in the simple case of an integer index
+        // with an odd-length list, return the index
+        return idx;
     }
```
mourner: @tmcw there's logic duplication between this function and `quantileSorted`, but deduplicating would mean either cutting out into a separate file to be used by both, or exporting it here but introducing a cycled dependency. What should we do?

mourner:
Pulled in `quickselect` to the repo. Keeping it private though because it's mutable. We'll expose it in the API in immutable form with `nthElement`.

src/quickselect.js
```diff
@@ -0,0 +1,70 @@
...
+ * var arr = [65, 28, 59, 33, 21, 56, 22, 95, 50, 12, 90, 53, 28, 77, 39];
+ * quickselect(arr, 8);
+ * // [39, 28, 28, 33, 21, 12, 22, 50, 53, 56, 59, 65, 90, 77, 95]
+ */
+function quickselect(arr /*: Array<number> */, k /*: number */, left /*: number */, right /*: number */) {
+    left = left || 0;
+    right = right || (arr.length - 1);
+
+    while (right > left) {
+        if (right - left > 600) {
```
tmcw: According to the Wikipedia pseudocode - https://en.wikipedia.org/wiki/Floyd%E2%80%93Rivest_algorithm - 600 is an 'arbitrary' constant' - could we comment / explain this more to make that fact more obvious. The same goes for 0.5

src/quickselect.js
```diff
@@ -0,0 +1,70 @@
...
+            var n = right - left + 1;
+            var m = k - left + 1;
+            var z = Math.log(n);
+            var s = 0.5 * Math.exp(2 * z / 3);
+            var sd = 0.5 * Math.sqrt(z * s * (n - s) / n);
```
tmcw: I can't find an open version of the original paper, but do these variables have longer names or meanings?

mourner: Nope, as you can see in the paper now :)

src/quickselect.js
mourner: Yes, added a comment.

tmcw:
:+1: awesome, thanks!

mourner:
@tmcw thanks for merging! Do you want to follow up with public `nthElement` API? I also didn't address https://github.com/simple-statistics/simple-statistics/pull/146#discussion_r57606181 but not sure if a clean solution exists.

PR #147, opened 2016-06-01
Use jsdoctest@^1.6.1 and fix the flow tests

yamadapc:
Adds `.+/node_modules/config-chain/test/.+` to the flow ignored files and uses
jsdoctest@^1.6.1 with _"inter-requiring modules have trouble with the
first-require trick"_ (yamadapc/jsdoctest#26) fixed.


tmcw:
👍 Awesome, thanks a ton!

PR #152, opened 2016-07-13
Implement bisect

eikooc:
find root of a function with the bisection method
https://en.wikipedia.org/wiki/Bisection_method

### DIFF
--- a/src/bisect.js
+++ b/src/bisect.js
@@ -15,7 +15,7 @@
  * @returns {Number} estimated root value
  * 
  * @example
- * bisect(Math.cos,0,4,100,0.003)
+ * bisect(Math.cos,0,4,100,0.003); // => 1.572265625
  */
 function bisect(f/*: (x: any) => number */, a/*: number */, b/*: number */, nMax/*: number */, tol/*: number */) {
     if (typeof f !== 'function') throw new SyntaxError('f must be a function');
--- a/src/bisect.js
+++ b/src/bisect.js
@@ -20,7 +20,7 @@ var sign = require('./sign');
  * bisect(Math.cos,0,4,100,0.003); // => 1.572265625
  */
 function bisect(func/*: (x: any) => number */, start/*: number */, end/*: number */, maxIterations/*: number */, errorTolerance/*: number */)/*:number*/ {
-    if (typeof f !== 'function') throw new TypeError('func must be a function');
+    if (typeof func !== 'function') throw new TypeError('func must be a function');
     
     var n = 1;
     while (n <= maxIterations) {
--- a/src/bisect.js
+++ b/src/bisect.js
@@ -17,7 +17,7 @@
  * @example
  * bisect(Math.cos,0,4,100,0.003)
  */
-function bisect(f, a, b, nMax, tol) {
+function bisect(f/*: (x: any) => number */, a/*: number */, b/*: number */, nMax/*: number */, tol/*: number */) {
     if (typeof f !== 'function') throw new SyntaxError('f must be a function');
     
     var n = 1;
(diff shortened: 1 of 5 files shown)


eikooc:
If you merge this pull request. Can you squash the commits with the following message:

> Implement bisect
> - find root of a function with the bisection method https://en.wikipedia.org/wiki/Bisection_method
>
> Implement sign function
> - This is required for bisect

src/bisect.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Number} a start of interval
+ * @param {Number} b end of interval
+ * @param {Number} nMax the maximum number of iterations
+ * @param {Number} tol the error tolerance
+ * @returns {Number} estimated root value
+ * 
+ * @example
+ * bisect(Math.cos,0,4,100,0.003); // => 1.572265625
+ */
+function bisect(f/*: (x: any) => number */, a/*: number */, b/*: number */, nMax/*: number */, tol/*: number */)/*:number*/ {
```
tmcw: since `a` is the start and `b` is the end, can you name the variables `start` and `end`? Preferring clarity over terseness is the pattern in simple statistics. Also:

* tol -> errorTolerance
* nMax -> maxIterations

src/sign.js
```diff
@@ -0,0 +1,29 @@
...
+    if (typeof x === 'number') {
+        if (x < 0) {
+            return -1;
+        } else if (x === 0) {
+            return 0
+        } else {
+            return 1;
+        }
+    } else {
+        throw new SyntaxError('not a number');
```
tmcw: a `TypeError` is likely more appropriate than a `SyntaxError` in this case.

src/bisect.js
```diff
@@ -0,0 +1,44 @@
...
+ * 
+ * This function returns a numerical approximation to the exact value.
+ * 
+ * @param {Function} f input function
+ * @param {Number} a start of interval
+ * @param {Number} b end of interval
+ * @param {Number} nMax the maximum number of iterations
+ * @param {Number} tol the error tolerance
+ * @returns {Number} estimated root value
+ *
```
tmcw: Since this method can throw an error, can you add a [throws tag](http://usejsdoc.org/tags-throws.html) to make that expectation clear to users?

tmcw:
This looks great! Functionality is solid, added some comments with suggestions around style and particulars. The other missing piece is re-exporting this function in `./index.js` so that module consumers can access it.

eikooc:
@tmcw it throws a standard `Error` and a `TypeError` I can't find how to document multiple `@throws` with JSDoc

src/bisect.js
```diff
@@ -0,0 +1,45 @@
...
+ * @throws {TypeError} Argument func must be a function
+ * 
+ * @example
+ * bisect(Math.cos,0,4,100,0.003); // => 1.572265625
+ */
+function bisect(func/*: (x: any) => number */, start/*: number */, end/*: number */, maxIterations/*: number */, errorTolerance/*: number */)/*:number*/ {
+    if (typeof func !== 'function') throw new TypeError('func must be a function');
+    
+    var n = 1;
+    while (n <= maxIterations) {
```
tmcw: could this be replaced by a traditional `for` loop?

```js
for (var i = 0; i <= maxIterations; i++) {
```

src/bisect.js
```diff
@@ -0,0 +1,45 @@
...
+
+var sign = require('./sign');
+/**
+ * [Bisection method](https://en.wikipedia.org/wiki/Bisection_method) is a root-finding 
+ * method that repeatedly bisects an interval to find the root.
+ * 
+ * This function returns a numerical approximation to the exact value.
+ * 
+ * @param {Function} f input function
+ * @param {Number} a start of interval
```
tmcw: the argument names in JavaScript are updated 👍  but the `@param` tags will also need to be updated

tmcw:
👍 awesome, thanks! Merging now.

PR #159, opened 2016-10-12
fix sample skewness equation

Xuefeng-Zhu:
It seems that correct equation is 
image

### DIFF
--- a/src/sample_skewness.js
+++ b/src/sample_skewness.js
@@ -31,7 +31,7 @@ function sampleSkewness(x /*: Array<number> */)/*:number*/ {
         cubedS = Math.pow(theSampleStandardDeviation, 3),
         sumCubedDeviations = sumNthPowerDeviations(x, 3);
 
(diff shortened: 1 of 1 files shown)


tmcw:
Can you fix the associated tests and confirm that this matches the output of another statistics library?

tmcw:
Comparing to [scipy.stats.skew](https://docs.scipy.org/doc/scipy-0.13.0/reference/generated/scipy.stats.skew.html),

* scipy returns `1.763632614803888`
* master returns `1.7636326148038879`
* this branch returns `7.0545304592155516`

The [R](https://cran.r-project.org/web/packages/moments/moments.pdf) skewness method returns a normalized value, and python does as well unless you specify Bias=false, which I did to generate this result.

---

I've confirmed that our existing implementation matches prior art, but also noticed that bias is expected rather than opt-in, so I think we should implement a `sampleSkewnessBiased` method to match Python and R.

Closing out here, since this change would make the output incorrect.

PR #160, opened 2016-10-13
Combinations with replacement

SJCaldwell:
Modified the algorithm I wrote to do combinations without replacement, but with replacement.

Turns out it was a pretty small fix.

Maybe it'd be preferable to just merge both of these into one file and have an argument for whether replacement should be considered or not? I'll leave that design choice up to you.

### DIFF
--- a/scripts/required_test.js
+++ b/scripts/required_test.js
@@ -37,6 +37,7 @@ require("../test/unique_count_sorted.test.js");
 require("../test/standard_deviation.test.js");
 require("../test/standard_normal_table.js");
 require("../test/sum.test.js");
+require("../test/combinationsReplacement.js");
 require("../test/sum_nth_power_deviations.test.js");
 require("../test/t_test.test.js");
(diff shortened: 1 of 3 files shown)


scripts/required_test.js
```diff
@@ -37,6 +37,7 @@ require("../test/unique_count_sorted.test.js");
 require("../test/standard_deviation.test.js");
 require("../test/standard_normal_table.js");
 require("../test/sum.test.js");
+require("../test/combinationsReplacement.js");
```
tmcw: Could this file be named `combinations_replacement.js`? Unfortunately with the intersection of OSX's case insensitivity and git's case sensitivity, uppercase characters in filenames can be very difficult to work with.

tmcw:
Merged with #165 - thanks @SJCaldwell! I've also added you to the contributors team so you'll be able to push feature branches to this repository directly.

PR #161, opened 2016-10-14
minor clean up bayesian_classifier

Xuefeng-Zhu:
minor clean up and remove unnecessary conditional statement

### DIFF
--- a/src/bayesian_classifier.js
+++ b/src/bayesian_classifier.js
@@ -57,7 +57,7 @@ BayesianClassifier.prototype.train = function(item, category) {
         }
 
         // And increment the key for this key/value combination.
-        this.data[category][k][item[k]]++;
+        this.data[category][k][v]++;
     }
 
     // Increment the number of items classified
@@ -83,7 +83,7 @@ BayesianClassifier.prototype.score = function(item) {
         for (category in this.data) {
             // Create an empty object for storing key - value combinations
             // for this category.
-            if (odds[category] === undefined) { odds[category] = {}; }
+            odds[category] = {};
 
             // If this item doesn't even have a property, it counts for nothing,
             // but if it does have the property that we're looking for from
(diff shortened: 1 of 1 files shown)


src/bayesian_classifier.js
```diff
@@ -104,10 +104,8 @@ BayesianClassifier.prototype.score = function(item) {
         // Tally all of the odds for each category-combination pair -
         // the non-existence of a category does not add anything to the
         // score.
+        oddsSums[category] = 0;
```
tmcw: Could you keep this change and remove the others?

src/bayesian_classifier.js
```diff
@@ -49,15 +49,15 @@ BayesianClassifier.prototype.train = function(item, category) {
         var v = item[k];
         // Initialize the nested object `data[category][k][item[k]]`
         // with an object of keys that equal 0.
-        if (this.data[category][k] === undefined) {
+        if (!this.data[category][k]) {
```
tmcw: I'd prefer the earlier syntax. What the `!` shortcut really does is

1. Cast the value to a boolean value
2. Compare that value to `false`

This has gotchas: `''` and 0 are values that cast to `false` but still _exist_. So I prefer `===` undefined as a uniform way to test for a lack of a value - otherwise you slip into the habit of using the `!` shortcut often and then introduce a bug where `0` is a legitimate value but is treated as a lack of value.

src/bayesian_classifier.js
```diff
@@ -83,7 +83,7 @@ BayesianClassifier.prototype.score = function(item) {
         for (category in this.data) {
             // Create an empty object for storing key - value combinations
             // for this category.
-            if (odds[category] === undefined) { odds[category] = {}; }
+            odds[category] = {};
```
tmcw: and also keep this change.

Xuefeng-Zhu:
@tmcw updated

tmcw:
Thanks @Xuefeng-Zhu!

Issue #157, opened 2016-09-22
Consider switching to faster Ckmeans implementation


schnerd:
It looks like Ckmeans support was added to simple-statistics by @tmcw in August 2015 as a part of #124. On May 25th, 2016 the original authors of the Ckmeans algorithm [released a new version](https://cran.r-project.org/web/packages/Ckmeans.1d.dp/NEWS) (3.4.6) that significantly increased to the performance of the core algorithm. The improvement is described in their [package overview](https://cran.r-project.org/web/packages/Ckmeans.1d.dp/Ckmeans.1d.dp.pdf), here's an excerpt of the relevant portion:

>A major speed-up has been introduced since version 3.4.6 using a divide-and-conquer strategy in dynamic programming. The runtime is reduced from O(kn^2) down to O(kn lg n), log-linear in
sample size n. This method is numerically stable and observed runtime is often close to linear in
practice. These improvements, not discussed in (Wang and Song, 2011), will be described in detail in a future publication.

Based on my inspection of the simple-statistics ckmeans implementation as well as performance tests, it definitely appears to be getting close to O(kn^2) runtime. Having an O(kn lg n) javascript implementation would be a major win. I wanted to open this issue to hear your thoughts on whether this improvement has already been discovered/discussed, and if you think it would make a good addition to simple-statistics.

tmcw:
Absolutely, the Ckmeans implementation in simple-statistics would definitely benefit from a faster algorithm, and if it's numerically stable, all the better. Is this something you'd be interested in porting to JS?

schnerd:
Glad to hear there's interest! I'm traveling over the next two weeks but I can probably take a stab at porting the code over when I return to the states. Will keep you updated!

schnerd:
Quick update–successfully ported the code and can confirm major speed improvements. I'm going to clean it up a bit more and open a PR.

schnerd:
Pull request open #163

tmcw:
Implemented in #163 - thanks @schnerd !

PR #163, opened 2016-10-16
Upgrade to faster Ckmeans implementation (Ckmeans 3.4.6)

schnerd:
See #157 for more details.

Ran some perf tests and the results are impressive:
screen shot 2016-10-13 at 10 12 35 am

While the code is functionally correct, I'm not 100% satisfied with the clarity of certain comments and variable/function names in this PR (in particular `ssq`, `sjlowi`, etc). The code from the original C++ implementation is unfortunately a bit difficult to follow, and the authors have not yet published a detailed explanation of the new algorithm, so it's difficult to design better names without a deeper understanding. 

I'm hoping to get a few more eyeballs on this code that can provide feedback or recommend changes if necessary.


src/ckmeans.js
```diff
@@ -27,6 +27,149 @@ function makeMatrix(columns, rows) {
...
+ */
+function ssq(j, i, sums, sumsOfSquares) {
+    var sji; // s(j, i)
+    if (j > 0) {
+        var muji = (sums[i] - sums[j - 1]) / (i - j + 1); // mu(j, i)
+        sji = sumsOfSquares[i] - sumsOfSquares[j - 1] - (i - j + 1) * muji * muji;
+    } else {
+        sji = sumsOfSquares[i] - sums[i] * sums[i] / (i + 1);
+    }
+    return Math.max(0, sji);
```
schnerd: I originally had this as
```js
if (sji < 0) {
    return 0;
}
return sji;
```
however the test coverage checker complained that the conditional return was not covered by any tests. I unfortunately was unable to craft a test case that hit this branch of the code, so I cheated by turning it into a Math.max and tricking the test coverage tool. A deeper understanding of the algorithm would be required to figure out when this conditional is needed or if it's even needed at all.

src/ckmeans.js
```diff
@@ -80,77 +226,16 @@ function ckmeans(data/*: Array<number> */, nClusters/*: number */)/*: Array<Arra
         return [sorted];
     }
 
-    // named 'D' originally
+    // named 'S' originally
     var matrix = makeMatrix(nClusters, sorted.length),
-        // named 'B' originally
+        // named 'J' originally
```
schnerd: These names were changed in the 3.4.6 implementation

src/ckmeans.js
```diff
@@ -80,77 +226,16 @@ function ckmeans(data/*: Array<number> */, nClusters/*: number */)/*: Array<Arra
...
-                        } else if (sumSquaredDistances + matrix[cluster - 1][j - 1] < matrix[cluster][sortedIdx]) {
-                            matrix[cluster][sortedIdx] = sumSquaredDistances + matrix[cluster - 1][j - 1];
-                            backtrackMatrix[cluster][sortedIdx] = j;
-                        }
-                    }
-                }
-            }
-        }
-    }
+    fillMatrices(sorted, matrix, backtrackMatrix);
```
schnerd: The body of `fillMatrices` could technically be inlined here. I kept it in a separate function since it's more true to the original implementation, but this can be easily changed if preferred.

tmcw: Whether this method is inline or not I'm not picky about - invocation overhead here is pretty minimal and it is nice to have shorter method bodies. The thing that doesn't fit that well with simple-statistics or JS is that these methods - fillMatrices and fillMatrixColumn - don't return values but change their input by reference. But reviewing the code, I don't see a good way to change this without reducing performance, so I think we're good with the implementation as-is.

mourner:
This looks wonderful, thanks for taking the time to port this!

mourner:
Can you rebase this branch so that it merges cleanly on the latest master?

src/ckmeans.js
llimllib: (If I had to guess I'd suspect it's to do with float math imprecisions?)

schnerd: @llimllib that's fantastic, can confirm that it works. will add it in with a comment.

schnerd:
@mourner suggestions taken into account and branch rebased on master

tmcw:
Awesome, thanks @schnerd - merging!

PR #164, opened 2016-10-18
update chunk example to make it more clear

Xuefeng-Zhu:
previous example is little ambiguous


tmcw:
👍 Thanks, yep, this makes it clearer

PR #172, opened 2017-02-02
Fix comment: value is undefined, not NaN

EvanHahn:
Something small that I noticed.


mourner:
Not sure why Circle didn't report the build but I can't merge because of that — want to push the green button @tmcw?

Issue #173, opened 2017-02-16
Implement random sample with replacement


jdittrich:
**Story:** As a user I would like to draw random samples with replacement to generate [bootstraps](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) (or, in general, generate a series of independent random values from an array).  

**Suggested implementation**: most natural would be adding an optional parameter to `shuffle` that allows to specify if the sample should be created with or without replacement.

tmcw:
Good idea! Random sampling with replacement, as far as I can tell, can be implemented as something like

```
for sample count ->
  pick random integer from 0 to size of dataset
  add dataset[random integer] to sample
```

Which is faster than an equivalent shuffle, since shuffles that sample the entire dataset would need to be O(n) wrt the dataset, whereas this can be O(n) or whatever, for the size of the sample.

PR #174, opened 2017-02-17
feat(core): sampleWithReplacement

tmcw:
Implements simple sampling with replacement, which, in contrast to the existing sample method,

allows an item to be chosen more than once.

Fixes https://github.com/simple-statistics/simple-statistics/issues/173

### DIFF
--- a/src/sample_with_replacement.js
+++ b/src/sample_with_replacement.js
@@ -1,0 +1,42 @@
+'use strict';
+/* @flow */
+
+/**
+ * Sampling with replacement is a type of sampling that allows the same
+ * item to be picked out of a population more than once.
+ *
+ * @param {Array} population an array of any kind of element
+ * @param {number} n count of how many elements to take
+ * @param {Function} [randomSource=Math.random] an optional entropy source that
+ * returns numbers between 0 inclusive and 1 exclusive: the range [0, 1)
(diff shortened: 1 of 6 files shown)


jdittrich:
Any support I could give? I started to write [something similar](https://github.com/jdittrich/simple-statistics/blob/sampleWithReplacement/src/sample.js) too, but your solution is probably more in line with the current code.

src/sample_with_replacement.js
```diff
@@ -0,0 +1,41 @@
...
+ * Sampling with replacement is a type of sampling that allows the same
+ * item to be picked out of a population more than once.
+ *
+ * @param {Array} population an array of any kind of element
+ * @param {number} n count of how many elements to take
+ * @param {Function} [randomSource=Math.random] an optional entropy source
+ * @return {Array} n sampled items from the population
+ * @example
+ * var sample = sampleWithReplacement([1, 2, 3, 4], 2);
+ * sampleWithReplacement; // = [2, 4] or any other random permutation
```
anandthakker: I think `permutation` implies sampling _without_ replacement, since a permutation is technically a rearrangement (so it couldn't contain duplicates).

tmcw:
Okay, got a quick review, this is good to go.

PR #182, opened 2017-02-20
Add to mean

Yomguithereal:
Adding a function taking the current mean, the number of elements in the list and a value to add and returning the new mean in constant time.

Will add more functions of the kind when I am confident I contributed the correct way.


tmcw:
Besides the errors I stupidly introduced when renaming parameters, this looks great! Merging.

Issue #177, opened 2017-02-18
Linear time mode computation


Yomguithereal:
Hello `simple-statistics`. Thanks for the awesome library.

I see that you chose to compute the mode by sorting the given array (which is cloned of course not to mutate the user's data) which runs in `O(n logn)` time.

Why not compute it in linear time by keeping track of the frequencies in linear time instead (it might be less fast because of JS engine optimizing `sort` to a certain extent but on large datasets, I doubt it).

tmcw:
Sure! There's a slightly better answer to this one :)

This design decision was because of objects in JavaScript: they're kind of a disappointment. The nicely linear implementation of mode computation would look like:

```js
var things = [true, 'true', false, 'false', false];

var counts = things.reduce((memo, thing) => {
  if (typeof memo[thing] == 'number') {
    memo[thing]++;
  } else {
    memo[thing] = 1;
  }
  return memo;
}, Object.create(null));

var modeCount = 0;
var mode = undefined;

Object.keys(counts).forEach(key => {
  if (counts[key] > modeCount) {
    mode = key;
    modeCount = counts[key];
  }
});

mode;
modeCount;
```

Live example: http://mistakes.io/#0bee75cd0fce87af1d530772f0e63938

Since objects in JavaScript can only have string keys, 0 and '0' will be counted the same, and false, 'false'.

The [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) object solves that problem by having arbitrary typed keys, but using it would require upping simple-statistics's browser requirements quite a bit (right now they're ~ES3).

There might be a linear implementation that I'm missing that doesn't require ES6 and works with strict equality - I'd love to make this faster if we can.

Yomguithereal:
Why not rely on some feature detection either when exporting the function or when applying it? Something like:

```js
typeof Map !== 'undefined';
```

Then we would have the best of both world, no? The `Map` object is quite optimized by recent browsers (even faster than plain objects in some use cases, notably this one).

But I agree this seems weird to expose different functions based on the engine's capabilities for  a library.

tmcw:
How about a `modeFast` method, and expose both? That'd dodge concerns like testability (we'd have to fake out Map availability), debuggability (bug reports would be clear about which implementation was at fault), as well as make the performance characteristics clearly documentable.

Yomguithereal:
[...]

PR #183, opened 2017-02-20
Implement indexed mode

tmcw:
Test for #177

I have some hacky benchmarks to push as well, which I should. It looks like modeFast is reliably faster than mode, but around 30%

### DIFF
--- a/src/mode_fast.js
+++ b/src/mode_fast.js
@@ -1,0 +1,54 @@
+'use strict';
+/* @flow */
+/* globals Map: false */
+
+/**
+ * The [mode](http://bit.ly/W5K4Yt) is the number that appears in a list the highest number of times.
+ * There can be multiple modes in a list: in the event of a tie, this
+ * algorithm will return the most recently seen mode.
+ *
+ * modeFast uses a Map object to keep track of the mode, instead of the approach
+ * used with `mode`, a sorted array. As a result, it is faster
+ * than `mode` and supports any data type that can be compared with `==`.
+ * It also requires a
+ * [JavaScript environment with support for Map](https://kangax.github.io/compat-table/es6/#test-Map),
+ * and will throw an error if Map is not available.
+ *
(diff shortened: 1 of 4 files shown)


src/mode_fast.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    var mode/*: number */;
+    var modeCount;
+
+    index.forEach(function (count, key) {
+        if (modeCount === undefined || modeCount < count) {
+            modeCount = count;
+            mode = key;
+        }
```
mourner: I think it would be simpler to move this logic to the loop above.

tmcw: Good idea, the benchmarks tell me it's faster too. Implemented in last commit.

src/mode_fast.js
```diff
@@ -0,0 +1,54 @@
...
+    // them like
+    // { value: count }
+    var index = new Map();
+
+    // A running `mode` and the number of times it has been encountered.
+    var mode;
+    var modeCount = 0;
+
+    for (var i = 0; i < x.length; i++) {
+        var newCount = index.get(x[i]);
```
tmcw: Perf note: was wondering if the lookup of `x[i]` these three times could be made faster by assigning `x[i]` to a variable. Tried it and benchmarks didn't show any evidence of a speedup.

src/mode_fast.js
```diff
@@ -0,0 +1,54 @@
...
+    // This index will reflect the incidence of different values, indexing
+    // them like
+    // { value: count }
+    var index = new Map();
+
+    // A running `mode` and the number of times it has been encountered.
+    var mode;
+    var modeCount = 0;
+
+    for (var i = 0; i < x.length; i++) {
```
tmcw: Perf note: was wondering if caching `x.length` would give a detectable speedup. Tried it, and it didn't.

Issue #194, opened 2017-04-10
Make it faster (master ticket)


tmcw:
![](https://media.tenor.co/images/83daa77b58d29d51e0b090dceeef7c96/tenor.gif)

In 3.0.0, I introduced [benchmarks](https://github.com/simple-statistics/simple-statistics/tree/master/benchmarks), so I now know how simple-statistics compares to other modules! Yay! And it 'wins' in some cases, but loses in others! Boo. I've done 30 solid minutes of digging here and can braindump a few results.

## Low-hanging fruit ✅ #195

`sumNthPowerDeviations` is internally called with n=2 and n=3. `Math.pow(x, 2)` is significantly slower than `x * x`: by swapping one for other, we'll get a quick win on sampleVariance:

```
global.ss.sampleVariance x 7,604,134 ops/sec ±0.81% (90 runs sampled)
global.ssMaster.sampleVariance x 9,283,292 ops/sec ±0.48% (92 runs sampled)
```

This is a bummer, since it isn't as 'literate' and makes the code slightly less similar to the written math algorithm, but alas, that's where we're at.

## Source of ambiguity

simple-statistics uses [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) by default, which is great since it's error-correcting and leads to more accurate sums and means. Unfortunately, it means extra work: it's slower than a normal sum, and since every other statistics library uses a simple sum algorithm instead, that's a performance hit.

I lean toward keeping Kahan summation as long as possible, but rethinking this decision sporadically.

## Design question

simple-statistics uses functions-calling functions so that algorithms can reuse each other: for instance, [sampleSkewness](https://github.com/simple-statistics/simple-statistics/blob/cfd98ade353336e2470015f0e6380ff7e22512ec/src/sample_skewness.js#L23-L36) uses sumNthPowerDeviations and sampleStandardDeviation under the hood. This is great for readability, but it is wasting some processor time, since, in that case, the call tree looks like:

* sampleSkewness
  * sumNthPowerDeviations (n=3)
    * **mean**
      * sum
  * sampleStandardDeviation
    * sampleVariance
      * sumNthPowerDeviations (n=2)
        * **mean**
          * sum

So, as you can see, we're computing **mean** twice. Now, mean is `O(n)`, but as I mentioned earlier, we're using Kahan sums, so it's a little more expensive than it might otherwise be, plus it feels _wrong_ as wasted work.

I can see a few solutions to this problem:

[...]

PR #195, opened 2017-04-10
perf(core): Improve performance of min, max, sumNthPowerDeviations, variance, sampleVariance

tmcw:
This captures low-hanging fruit optimizations: switching from Math.pow(n, 2) to n * n, and using the
first element of the array as the starting minimum and maximum value, allowing us to skip an
undefined check in each loop.

References https://github.com/simple-statistics/simple-statistics/issues/194


Yomguithereal:
You should also try to cache your arrays' length for your for loop. It's known to avoid needing to ask for it at each iteration.

tmcw:
As far as I can tell, [cached array length is no longer necessary in V8, at least](https://jsperf.com/caching-array-length/4), which covers Node and Chrome, and I'd assume that Safari & Firefox will eventually match that optimization. If there are tests that support a meaningful advantage against uncached array lengths, I'd definitely go for it.

PR #197, opened 2017-04-11
perf(sampleSkewness): Improve sampleSkewness performance

tmcw:
Avoids a recomputation of the mean value

Refs #194 

~This introduces one new internal method: sumNthPowerDeviationsFrom, which lets you specify the other number, instead of computing the mean. And in the case of sampleSkewness, instead of using a different method to compute standardDeviation, we compute it inline.~

I've changed this implementation to instead compute deviations n=2 and n=3 in the same loop, and to avoid calling a separate method at all, which makes it even faster - a solid 2.6x boost.

cc @mourner - think this is a good balance?


mourner:
@tmcw perfect. I think it's totally acceptable to write a direct implementation instead of composing public methods if it's still simple enough and comes with a big perf boost.

PR #200, opened 2017-04-12
Implementing the Kahan-Babuska algorithm

Yomguithereal:
Hello `simple-statistics`.

I tried some things to improve the performance of the `sum` function without compromising the float rounding error correction and stumbled upon the Kahan-Babuska algorithm (you can read a description of the algorithm in [this](https://pdfs.semanticscholar.org/1760/7d467cda1d0277ad272deb2113533131dc09.pdf) paper).

It's faster because it can apply the correction at the end rather than at each iteration.

What do you think?

### DIFF
--- a/src/sum.js
+++ b/src/sum.js
@@ -2,15 +2,17 @@
 /* @flow */
 
 /**
- * Our default sum is the [Kahan summation algorithm](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) is
- * a method for computing the sum of a list of numbers while correcting
- * for floating-point errors. Traditionally, sums are calculated as many
+ * Our default sum is the [Kahan-Babuska algorithm](https://pdfs.semanticscholar.org/1760/7d467cda1d0277ad272deb2113533131dc09.pdf).
+ * This method is an improvement over the classical
+ * [Kahan summation algorithm](https://en.wikipedia.org/wiki/Kahan_summation_algorithm).
+ * It aims at computing the sum of a list of numbers while correcting for
+ * floating-point errors. Traditionally, sums are calculated as many
  * successive additions, each one with its own floating-point roundoff. These
  * losses in precision add up as the number of numbers increases. This alternative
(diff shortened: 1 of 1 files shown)


Yomguithereal:
Refs #194.

tmcw:
This is fantastic! Going to merge and then add `sum` to the benchmark table.

Also, in the spirit of [open open source](http://openopensource.org/), I've added you as a contributor to this repository! Should make contributing a bit simpler. Thanks for this great improvement.

tmcw:
Minor note: to auto-generate changelogs, I've been using the [conventionalcommits](https://conventionalcommits.org/) convention for messages - either manually when I squash commits (like here) or with `git cz`, which gives a nice UI for choosing messages. This way changelogs can be auto-generated in new versions.

Yomguithereal:
Thanks @tmcw.

> Minor note: to auto-generate changelogs, I've been using the conventionalcommits convention for messages - either manually when I squash commits (like here) or with git cz, which gives a nice UI for choosing messages. This way changelogs can be auto-generated in new versions.

I'll try to follow the convention now.

Concerning the Kahan-Babuska algorithm, I'd like if someone could double-check my code & the algorithm to be sure we are not doing something wrong.

I also noticed that the error correction was not performed with the `sumNthPowerDeviations`.

tmcw:
Yeah: so right now the gist is, for `sum()`, we do error correction, but there are other places where we compute sums 'inline', where we don't. To reuse the error correcting properties of the `sum()` algorithm, we'd have to collect a list of all the numbers we want to sum and then sum it, whereas right now we're doing it online by simply adding.

We could

* Do error correction within each method that sums, re-implementing this algorithm each time
* Collect the full list of results and make sure that isn't a performance problem
* Use a generator function? That'd really increase the minimum JavaScript target for simple-statistics, unfortunately

The Kahan-Babuska algorithm checks out - it's almost the pseudocode verbatim, and gives the same precision as in the previous Kahan implementation.

PR #205, opened 2017-04-14
Faster binomialDistribution

mourner:
Improves performance of `binomialDistribution` by computing binomial coefficient cumulatively instead of using factorials in each iteration. Ref #194

Improves time from 3.8s to 450ms for the following simple benchmark:

```js
for (var i = 1; i < 50000; i++) binomialDistribution(i, Math.random());
```

### DIFF
--- a/src/binomial_distribution.js
+++ b/src/binomial_distribution.js
@@ -2,7 +2,6 @@
 /* @flow */
 
 var epsilon = require('./epsilon');
-var factorial = require('./factorial');
 
 /**
  * The [Binomial Distribution](http://en.wikipedia.org/wiki/Binomial_distribution) is the discrete probability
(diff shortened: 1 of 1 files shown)


mourner:
@tmcw while doing #207, I noticed that both `binomialDistribution` and `bernoulliDistribution` use Object as the return value, although functionally it is fully equivalent to Array because all keys are sequential integers starting from 0. 

If I switch the implementation to array (simply `cells = []` instead of `cells = {}`), it runs 25% faster. Should we switch?

tmcw:
> If I switch the implementation to array (simply cells = [] instead of cells = {}), it runs 25% faster. Should we switch?

👍 Yep. And I'll be semver-strict about that and release a 4.0.0 - that'll also let us cover the removal of factorial.

mourner:
@tmcw OK, let's follow-up with `[]` change after this and #207 are merged.

PR #208, opened 2017-04-14
Use Math.exp instead of Math.pow

mourner:
A further ~5% poissonDistribution performance improvement.


tmcw:
👍 Perfect. Checked and Math.exp has been around since ECMAScript 1, so we're all good compatibility-wise.

PR #209, opened 2017-04-14
Return array instead of object in distribution functions

mourner:
Makes `bernoulliDistribution`, `binomialDistribution` and `poissonDistribution` return an array instead of object. The result is fully requivalent since keys were sequential integers starting from 0, but this is ~5% faster.


tmcw:
Can you add [breaking change](https://conventionalcommits.org/) to the commit message, as spec'ed by Conventional Commits, so that standard-changelog/standard-version will properly mark it as such?

PR #212, opened 2017-04-16
ES6 Modules

### LINKED ISSUE #189: Use prettier
I'm going prettier-by-default across all my projects, and this is one of them.

### LINKED ISSUE #210: Use ES6 modules + Rollup
As the front-end world rapidly moves to ES6-compatible bundling, simple-statistics could really benefit from switching from CommonJS to ES6 modules and Rollup. It may introduce small additional set up for tests and NPM prepublish scripts, but advantages now overweight the cons IMO. The library will especially benefit from tree-shaking (also see #180), leaner browser builds, and being future-proof.

SS would still have a CommonJS-compatible entry point for Node, but additionally use `pkg.module` for ES6-aware consumers. The Webpack code splitting issues should not hold us back because they're not ES6-specific, and are probably still a problem when you try to set up code splitting with CommonJS modules. 

Examples of other libraries that switched happily to Rollup include React, Vue, Ember, Preact, D3, Three.js, Moment and PouchDB.

cc @tmcw @Yomguithereal

tmcw:
* Fixes #189 
* Fixes #210

### Working!

* Build script that bakes CommonJS and ES6 entry points with Rollup, strips Flow
* Tests that pass

### Not working yet, questions...

* Test coverage doesn't work. node-tap / nyc supports sourcemaps but it isn't working yet. Maybe they need to be inline?
* Should tests hit the bundle (like right now) or ES6 source?
* jsdoctest doesn't work yet.
* Since tests rely on a CommonJS bundle, they didn't have access to internals like numeric_sort or sign. So I exposed these modules, and would still prefer not to. Maybe tests shouldn't rely on the bundle?

cc @mourner for thoughts


tmcw:
I'm going to wait until ES6 modules are natively supported in Node, or there's a breakthrough in testing for ES6. As tech currently stands, I can't find any option that is

1. Simple
2. Supports coverage
3. Stacktraces work
4. You don't have to require a compiled file

rowanwins:
Gday @tmcw 

I'm interested in this and wondering I can help here? I'm currently using ss in a library I'm putting together and would love to get a slimmer build.

For my libary Im using rollup and have been pairing it with [ava](https://github.com/avajs/ava) for es6 testing, I've been pretty happy with it. 
1. It's simple, pretty much a drop in replacement for tap
2. Ava supports [code coverage](https://github.com/avajs/ava/blob/master/docs/recipes/code-coverage.md) via nyc.
3. Stacktraces work 
4. You don't have to require compiled files

If you're interested in this as an option I'm happy to take a look and see if it I can get it working...

tmcw:
ava might be an option: my fear around it and other Babel-based approaches is that I _don't_ want to use Babel for building dists, and yet ava's [default ES2017 support](https://github.com/avajs/ava#es2017-support) covers more than just `import/export`. Which means that simple-statistics could start using features that require transpiling, tests will pass, but we won't be transpiling those features in the distribution yet.

ava with a custom babel configuration that only includes the features supported in rollup + ES5 would work - I haven't tinkered with that idea yet and it's definitely worth a look.

rowanwins:
Thanks for sharing those areas for investigation. I'll have a quick play and see if I can get Ava playing nicely with the rollup Babel configuration and see where it lands and report back

rowanwins:
Hmmm @tmcw you were right about this being complicated :)

So I've got things working partially
````
  "ava": {
    "require": [
      "babel-register"
    ],
    "babel": "inherit"
  },
  "babel": {
    "presets":[],
    "plugins": [
      ["transform-es2015-modules-commonjs", {
          "strict": true,
          "noInterop": true
      }]
    ]
},
````
And then in bayes.test.js
````
import { BayesianClassifier } from '../src/bayesian_classifier'
import test from 'ava';

test('BayesianClassifier', function(t) {
... etc
````
With some minor modifications to the tests they work

**However** with the current config Babel and/or Ava seems to be transpiling other ES5 code if I include it in the src. So for example in bayesian_classifier.js I included a [default value](https://googlechrome.github.io/samples/default-parameters-es6/) on the `train` method, the test picked this up and passed (although rollup wouldn't build it). I'm not enough of an expert on Babel to work out why this might be happening...

The only other thought I've had is using something like the [rollup-plugin-babel](https://github.com/rollup/rollup-plugin-babel) and the [babel-preset-es2015-rollup](https://github.com/rollup/babel-preset-es2015-rollup). We might be able to get everything a bit more in sync then.

rowanwins:
Looking at the [ava doco](https://github.com/avajs/ava#transpiling-imported-modules) a bit further it talks about how the babel settings are only used on the test files, rather than the files they import

In the ava examples they tend to use 
````
  "ava": {
    "require": [
      "babel-register"
    ],
````
I gather the `require` key is a babel hook so Im wondering if this is where we need to pass in `transform-es2015-modules-commonjs` somehow... 

Anyway im pretty confused, time to do something completely different and bake a cake :)

PR #217, opened 2017-04-27
Fixes JSDoc in sampleSkewness

emanuelfeld:
Edited to reflect error throwing when input array has length <3.


tmcw:
Thanks Emanuel!

Issue #244, opened 2017-09-17
Kernel density estimation


anandthakker:
[Kernel density estimation](https://en.wikipedia.org/wiki/Kernel_density_estimation) is a useful tool for (among other things) estimating the shape of the underlying probability distribution from a sample.  @tmcw would you be interested in / open to a PR adding KDE functionality?  Something like:

```js
kde: (x: Array<number>, kernel: KernelFunction, bandwidth: number) => DensityFunction
```

`KernelFunction` is a `number => number` that's non-negative and integrates to 1.

The hard part of using KDE (for non-experts like me, anyway) is the selection of the kernel and bandwidth -- indeed, that's much harder than the implementation of `kde` itself -- so I think the real value of having something like this in `simple-statistics` would be in providing good defaults for those:
 - [R](https://www.rdocumentation.org/packages/stats/versions/3.4.1/topics/bandwidth) defaults to a Gaussian kernel and uses what appears to be a common/traditional [rule-of-thumb bandwidth](https://en.wikipedia.org/wiki/Kernel_density_estimation#A_rule-of-thumb_bandwidth_estimator) choice.
 - [Haskell's `Statistics.Sample.KernelDensity`](https://hackage.haskell.org/package/statistics-0.14.0.2/docs/Statistics-Sample-KernelDensity.html) uses a different method, from [this paper by Botev, et al](https://arxiv.org/pdf/1011.2602.pdf) which the docs claim outperform the above approach.  (Haven't yet read the paper.)

anandthakker:
cc @jfirebaugh

tmcw:
👍 Yep, would definitely be open to a PR! Other prior arts are [scipy, which explicitly calls it 'gaussian kde'](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html), and [science.js, which sets gaussian to the default and lets you customize it](https://github.com/jasondavies/science.js/blob/master/src/stats/kde.js).

PR #245, opened 2017-09-18
Add kernel density estimation

anandthakker:
Refs #244


anandthakker:
@tmcw quick draft to get any early thoughts on how this should end up looking

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,88 @@
...
+ * @param bandwidth The "bandwidth selection" method to use.  If a function is provided, it should compute a positive bandwidth value from the sample.
+ * @returns An estimated [probability density function](https://en.wikipedia.org/wiki/Probability_density_function) for the given sample. The returned function runs in `O(X.length)`.
+ */
+function kde(
+    X /*: Array<number> */,
+    kernel /*: $Keys<typeof kernels> | ((number) => number) | void*/,
+    bandwidth /*: $Keys<typeof bandwidthMethods> | number | void*/
+) {
+    var kernelFn/*: (number) => number */;
+    if (typeof kernel === 'undefined') {
```
tmcw: `kernel === undefined` should work here, there's no risk of a ReferenceError when the binding comes in as an argument.

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,88 @@
...
+ * Well known bandwidth selection methods
+ * @private
+ */
+var bandwidthMethods /*: {[string]: (Array<number>) => number} */ = {
+    /**
+     * The commonly used variation of [Silverman's
+     * rule-of-thumb](https://en.wikipedia.org/wiki/Kernel_density_estimation#A_rule-of-thumb_bandwidth_estimator) found in Scott, D. W. (1992) Multivariate Density Estimation: Theory, Practice, and Visualization.
+     * @private
+     */
+    nrd: function (x /*: Array<number> */) {
```
tmcw: What is `nrd` an abbreviation of? I'm not seeing anything that refers to NRD in the linked Wikipedia page.

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,88 @@
...
+    } else if (typeof bandwidth === 'string') {
+        if (!bandwidthMethods[bandwidth]) {
+            throw new Error('Unknown bandwidth method "' + bandwidth + '"');
+        }
+        bw = bandwidthMethods[bandwidth](X);
+    } else if (typeof bandwidth === 'number') {
+        bw = bandwidth;
+    }
+
+    var N = X.length;
```
tmcw: It looks like X is never mutated, so this is a perf optimization. Could we avoid this optimization? I haven't been able to establish evidence of an across-the-board advantage or even consistent advantage for that technique.

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * [Kernel density estimation](https://en.wikipedia.org/wiki/Kernel_density_estimation)
+ * is a useful tool for, among other things, estimating the shape of the
+ * underlying probability distribution from a sample.
+ *
+ * @param X sample values
+ * @param kernel The kernel function to use. If a function is provided, it should return non-negative values and integrate to 1. Defaults to 'gaussian'.
+ * @param bandwidth The "bandwidth selection" method to use.  If a function is provided, it should compute a positive bandwidth value from the sample.
+ * @returns An estimated [probability density function](https://en.wikipedia.org/wiki/Probability_density_function) for the given sample. The returned function runs in `O(X.length)`.
```
tmcw: The returns tag should have a (Function) type

src/kernel_density_estimation.js
anandthakker: Oops, yeah this comes from https://stat.ethz.ch/R-manual/R-devel/library/stats/html/bandwidth.html, and it's also used in `science.js`, but I've had no luck figuring out what it means.  What do you think is better:
 - Change the link from Wikipedia to the R documentation
 - Rename from `nrd` to `scott`, and add a link to the scipy docs that explicitly distinguish Silverman from Scott (they're basically the same modulo a constant, but the Wikipedia page doesn't get into it)
 - Just call it `silverman`, because it's searchable on the wikipedia page.

anandthakker: Scipy link: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html

jfirebaugh: "nrd" appears to stand for "Normal Reference Distribution", based on https://stat.ethz.ch/R-manual/R-devel/library/MASS/html/bandwidth.nrd.html.

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,94 @@
...
+var bandwidthMethods /*: {[string]: (Array<number>) => number} */ = {
+    /**
+     * The ["normal reference distribution"
+     * rule-of-thumb](https://stat.ethz.ch/R-manual/R-devel/library/MASS/html/bandwidth.nrd.html),
+     * a commonly used version of [Silverman's
+     * rule-of-thumb](https://en.wikipedia.org/wiki/Kernel_density_estimation#A_rule-of-thumb_bandwidth_estimator).
+     * @private
+     */
+    nrd: function (x /*: Array<number> */) {
+        var h = stddev(x);
```
tmcw: From the wikipedia page about this, it seems that "H" is a smoothing parameter called the bandwidth. Can we call it "bandwidth" instead of H?

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,94 @@
...
+    } else if (typeof kernel === 'string') {
+        if (!kernels[kernel]) {
+            throw new Error('Unknown kernel "' + kernel + '"');
+        }
+        kernelFn = kernels[kernel];
+    } else {
+        kernelFn = kernel;
+    }
+
+    var bw/*: number */;
```
tmcw: What does `bw` represent? Why is it separate from `bandwidth`?

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param bandwidth The "bandwidth selection" method to use.  If a function is provided, it should compute a positive bandwidth value from the sample.
+ * @returns {Function} An estimated [probability density function](https://en.wikipedia.org/wiki/Probability_density_function) for the given sample. The returned function runs in `O(X.length)`.
+ */
+function kde(
+    X /*: Array<number> */,
+    kernel /*: $Keys<typeof kernels> | ((number) => number) | void*/,
+    bandwidth /*: $Keys<typeof bandwidthMethods> | number | void*/
+) {
+    var kernelFn/*: (number) => number */;
+    if (kernel === 'undefined') {
```
tmcw: should be `kernel === undefined`, without the stringyquotes

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,94 @@
...
+            throw new Error('Unknown bandwidth method "' + bandwidth + '"');
+        }
+        bw = bandwidthMethods[bandwidth](X);
+    } else if (typeof bandwidth === 'number') {
+        bw = bandwidth;
+    }
+
+    return function (x /*: number*/) {
+        var i = 0;
+        var output = 0;
```
tmcw: is there a better name for `output` here, considering that it isn't actually the output?

src/kernel_density_estimation.js
anandthakker: Ah -- I think the entire quantity being returned by this function is the smoothing parameter / "bandwidth".  `h` is just an intermediate value in the computation.  If we wanted to give it a meaningful name rather than `h`, maybe... `spread`?

src/kernel_density_estimation.js
anandthakker: whoops, thanks

src/kernel_density_estimation.js
anandthakker: Although, tbh I think calling it `spread` imbues it with more suggested meaning than it should have.

src/kernel_density_estimation.js
anandthakker: Eh, this was just to satisfy flow.  What I wanted to do was just reassign it in the case that the given `bandwidth` was a named method (or omitted):

```js
    if (typeof bandwidth === 'undefined') {
        bandwidth = bandwidthMethods.nrd(X);
    } else if (typeof bandwidth === 'string') {
        if (!bandwidthMethods[bandwidth]) {
            throw new Error('Unknown bandwidth method "' + bandwidth + '"');
        }
        bandwidth = bandwidthMethods[bandwidth](X);
    }
```

src/kernel_density_estimation.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * [Kernel density estimation](https://en.wikipedia.org/wiki/Kernel_density_estimation)
+ * is a useful tool for, among other things, estimating the shape of the
+ * underlying probability distribution from a sample.
+ *
+ * @param X sample values
+ * @param kernel The kernel function to use. If a function is provided, it should return non-negative values and integrate to 1. Defaults to 'gaussian'.
+ * @param bandwidthMethod The "bandwidth selection" method to use, or a fixed bandwidth value. Defaults to "nrd", the commonly-used ["normal reference distribution" rule-of-thumb](https://stat.ethz.ch/R-manual/R-devel/library/MASS/html/bandwidth.nrd.html).
```
anandthakker: Duplicated the description/link for normal reference distribution here since, as of now, the `bandwidthMethods` object is private.

src/kernel_density_estimation.js
anandthakker: I went with just `s`, so as not to confusingly collide with `h` on the wikipedia page, but also not promise any especially interpretable meaning for the temporary quantity.

anandthakker:
@tmcw updated to address the variable name clarity problems

PR #246, opened 2017-09-23
Implement ES6 modules - Fixes #210

### LINKED ISSUE #210: Use ES6 modules + Rollup

DenisCarriere:
## Implementation of ES6 modules 💥 🎉 

Ref: https://github.com/simple-statistics/simple-statistics/issues/210

- [x] Include Rollup configuration to `dist/simple-statistics.js`
- [x] Add source mapping & minify directly in Rollup config (1.5s to build Rollup bundle)
- [x] Documentation JSDocs test
- [x] Replaced Browserify + minify with Rollup
- [x] Convert all source code to ES6 export/import
- [x] Convert tests to use main library `var ss = require('../')`
- [x] Dropped `"use strict"` (unnecessary inside of modules)
- [x] Fix JSDocs linting issues
- [x] Dropped `.npmignore` (not required if you include `files` in `package.json`)

Tons of changes, but it's really not too bad to review, I've made sure to only change what's needed to make the ES6 module implementation.

Rollup config & module export syntax was inspired by @mbostock's [`d3-geo-projection`](https://github.com/d3/d3-geo-projection).

CC: @mourner @tmcw

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,10 @@ module.exports = {
         "node": true
     },
     "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": 6,
+        "sourceType": "module"
+    },
     "rules": {
         "accessor-pairs": "error",
         "array-bracket-spacing": [
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,4 +1,5 @@
 [ignore]
+.*/node_modules/documentation/.*
 .*/node_modules/config-chain/test/.*
 .*/node_modules/conventional-changelog-core/test/.*
 .*/node_modules/.cache.*
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,0 @@
-coverage
-test
-docs
-benchmarks
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ in all modern browsers (including IE) as well as in [node.js](https://nodejs.org
   * When you use simple-statistics from a script tag, you don't get to choose
     the variable name it is assigned to: simple-statistics will always become
     available globally as the variable `ss`. You can reassign this variable to
-    another name if you want to, but doing so is optional. <pre><script src='https://unpkg.com/simple-statistics@4.1.0/dist/simple-statistics.js' /></pre>
+    another name if you want to, but doing so is optional. <pre><script src='https://unpkg.com/simple-statistics@4.1.1/dist/simple-statistics.js' /></pre>
     There are two options for the `src` attribute of that script tag: one with
     `.min.js` that is compressed, and the other without, that is raw.
-    * `https://unpkg.com/simple-statistics@4.1.0/dist/simple-statistics.js`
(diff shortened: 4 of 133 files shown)


tmcw:
Wow, awesome :) This would probably be a 5.x, because many simple-statistics users use `require('simple-statistics/min')` and other sub-requires to get just one module, whereas now they will always import `simple-statistics` but possibly only some of its bindings.

I think the only thing to nail down here is the removal of jsdoctest - that project was unfortunately [stalled for a while, but has a new release](https://github.com/yamadapc/jsdoctest/issues/33), but does this mean that... it wouldn't support ES6?

DenisCarriere:
👍 Thanks 😄 

> This would probably be a 5.x

Agreed, changing the library entirely to ES6 modules will break some minor things such `require('simple-statistics/min')` since `min.js` will now contain `export default`.

**Before**

```js
const min = require('simple-statistics/min');
```

**Now ES6**

```js
import { min } from 'simple-statistics';
```

**Now CommonJS**

```js
const min = require('simple-statistics').min;
```

> I think the only thing to nail down here is the removal of jsdoctest 

Ops! Didn't realize I removed `jsdoctest`, I'll look into re-implementing it, I don't see why it shouldn't work... 🤔

DenisCarriere:
# JSDoctest Update

Got it to work using `mocha --require jsdoctest dist/index.js`, I simply added a CommonJS build to the Rollup config. Currently there's 3 Rollup outputs (UMD, UMD + minified & CommonJS) all with sourcemaps.

DenisCarriere:
## Rollup Config

Here's the latest `rollup.config.js`, made it more "compact" by adding default settings and it's now way easier to read/understand.

```js
import uglify from 'rollup-plugin-uglify';

function assign({ file, format, plugins }) {
    return {
        input: 'index',
        output: {
            file,
            format: format || 'umd',
            extend: true,
            sourcemap: true,
            name: 'ss'
        },
        plugins: plugins || []
    }
}
export default [
    assign({ file: 'dist/simple-statistics.js' }),
    assign({ file: 'dist/simple-statistics.min.js', plugins: [uglify()] }),
    assign({ format: 'cjs', file: 'dist/index.js' })
];
```

DenisCarriere:
👍 🎉 Woot!

@tmcw Feel free to refactor it to your likings, many options are viable.

PR #252, opened 2017-10-04
chore(benchmark): benchmark upgrade

Yomguithereal:
* upgrading ssMaster version
* differentiating between `variance` & `sampleVariance` in the benchmark


Yomguithereal:
I haven't rewritten the readme though because the benchmark is yielding strange result between ss and ssMaster in regards to variance performance.

Issue #253, opened 2017-10-13
Missing index.js for module support


DenisCarriere:
Just noticed that `index.js` wasn't being published with the latest npm release `v5.0`.

![image](https://user-images.githubusercontent.com/550895/31552063-64c853b4-b004-11e7-8cc8-3b563192e135.png)

Easy fix would be to include `index.js` in the **files**

**package.json**
```json
  "files": [
    "bower.json",
    "src",
    "dist",
    "index.js",
  ],
```

PR #254, opened 2017-10-13
Add index.js to package.json

DenisCarriere:
Ref: https://github.com/simple-statistics/simple-statistics/issues/253


DenisCarriere:
I guess not, I thought so too!

PR #255, opened 2017-10-13
Implementing Typescript

DenisCarriere:
# Implementing Typescript (proposal)

To make it easier to maintain, the Typescript definition is almost an exact copy of the origin Javascript files, the main `index.js` export remains the same and all the methods are simply copying the function definition.

## Things to think about

- Include the JSDocs in the typescript definitions that way VSCode's [`intellisense`](https://code.visualstudio.com/docs/editor/intellisense) will still capture all the `@example` & descriptions.
- Automating JSDocs updates to Typescript definition (that way you only need to maintain the Javascript files and not worry about the Typescript definitions)
- Better to have multiple `d.ts` files for each module since some of the modules are Classes and those can get quite complex instead of pure functions. Also having multiple `.d.ts` files will be easier to automate update JSDocs.
- Added `test/types.ts` to make sure the Typescript definition are properly working, we could eventually automate this to pull this down from the `@example` in the JSDocs. **Note:** this doesn't actually test if the method works or not, it only tests if the Typescript definition works properly for it's input/output.

## Implemented the following definitions:

- [x] addToMean
- [x] combineMeans
- [x] combineVariances
- [x] geometricMean
- [x] harmonicMean
- [x] average
- [x] median
- [x] medianSorted
- [x] subtractFromMean
- [x] A lot more to do...


DenisCarriere:
@tmcw Let me know if you want me to go ahead using this approach, I'll finish up the rest of the methods.

This should be relatively easy to maintain since these methods won't be change (unless there's a major release), so once the definitions are created, you can forget about them for a while. Just `@` me if you have any questions on the complex class based ones.

src/add_to_mean.d.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @since 2.5.0
+ * @param {number} mean current mean
+ * @param {number} n number of items in the list
+ * @param {number} newValue the added value
+ * @returns {number} the new mean
+ *
+ * @example
+ * addToMean(14, 5, 53); // => 20.5
+ */
+declare function addToMean(mean: number, n: number, newValue: number): number
```
tmcw: Do typescript definitions have to repeat the documentation strings?

DenisCarriere: 🤔 Well in practice no... but then you'd lose the JSDocs information provided such as the `@since`, `@example` & the description.

## With JSDocs

image

## Without JSDocs

image

DenisCarriere:
@tmcw Your choice for Typescript definition.

## Option 1. Very simplistic

> Easiest one to maintain, no need to update JSDocs, however less information provided to the user about the method used and no examples provided.

```ts
function addToMean(mean: number, n: number, newValue: number): number;

export default addToMean;
```

## Option 2. With JSDocs

> Still easy to maintain, the same level of effort as Option 1, however JSDocs might get outdated if edits are made in the normal Javascript files.

```ts
/**
 * When adding a new value to a list, one does not have to necessary
 * recompute the mean of the list in linear time. They can instead use
 * this function to compute the new mean by providing the current mean,
 * the number of elements in the list that produced it and the new
 * value to add.
 *
 * @since 2.5.0
 * @example
 * addToMean(14, 5, 53); // => 20.5
 */
declare function addToMean(mean: number, n: number, newValue: number): number;

export default addToMean;
```
## Option 3. Compact

> All the definitions in a single `index.d.ts` file.
> I personally think this would be difficult to maintain.

**index.d.ts**

```ts
export function addToMean(mean: number, n: number, newValue: number): number;
export function combineMeans(mean1 : number, n1: number, mean2: number, n2: number): number;
export function geometricMean(x: Array<number>): number;
```

DenisCarriere:
## Option 4. Minimal + URL as reference

```ts
/**
 * https://simplestatistics.org/docs/#addtomean
 */
declare function addToMean(mean: number, n: number, newValue: number): number;

export default addToMean;
```

CC: @tmcw I prefer the website Option 4.

DenisCarriere:
@tmcw Ready to merge, feel free to suggest changes.

This PR even fixes some Flow types.

CjS77:
+1 to merge this PR please

DenisCarriere:
@CjS77 If you are pulling the branch directly, you'll need to run `npm run build` to create the CommonJS files to import the module directly from the folder.

DenisCarriere:
@tmcw Going to merge, let us know whenever a new release is published to test this on the live npm release.

tmcw:
👍 5.1.0 is out with these changes included.

DenisCarriere:
🎉 woot!

Issue #203, opened 2017-04-13
Standard error


Yomguithereal:
Correct me if I am wrong but I think we don't have a standard error function.

tmcw:
We don't! We have the unrelated gaussian error function, but not standard error. This'd be a great addition.

PR #257, opened 2017-10-16
feat: standardError

Yomguithereal:
Adding the `standardError` function to the library.

Some interrogations:

* What should we return when the list has only 1 element? `0` as done now?
* What should we do with an empty list? Throw? (Note that this case is not addressed with the `standardDeviation` function either).
* Should we also add `sampleStandardError` with Bessel's correction?

Note that I can add some unit tests to reflect our choices on the first two questions.

### DIFF
--- a/src/standard_error.js
+++ b/src/standard_error.js
@@ -1,0 +1,25 @@
+/* @flow */
+
+import standardDeviation from './standard_deviation';
+
+/**
+ * The [standard error](https://en.wikipedia.org/wiki/Standard_error)
+ * of a statistic (here the mean) is the standard deviation of its sampling
+ * distribution or sometimes an estimate of that standard deviation.
+ *
+ * @param {Array<number>} x input
+ * @returns {number} standard error
+ * @example
+ * standardError([2, 4, 4, 4, 5, 5, 7, 9]); // => 2 / Math.sqrt(8)
+ */
+function standardError(x /*: Array<number> */)/*:number*/ {
+    if (x.length === 1) {
+        return 0;
+    }
+
+    var std = standardDeviation(x);
+
+    return std / Math.sqrt(x.length);
+}
+
+export default standardError;
--- a/src/standard_error.js
+++ b/src/standard_error.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import standardDeviation from './standard_deviation';
+import standardDeviation from "./standard_deviation";
 
 /**
  * The [standard error](https://en.wikipedia.org/wiki/Standard_error)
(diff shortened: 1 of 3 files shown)


Yomguithereal:
Related to #203.

Yomguithereal:
Nobody wants standard error :(. Sadness fills my heart.

Yomguithereal:
Bump?

jwilber:
For your first two interrogations,
```
What should we return when the list has only 1 element? 0 as done now?
What should we do with an empty list? Throw? (Note that this case is not addressed with the standardDeviation function either).
```

I vote for both we return `nan `.
I'd have to review, but I believe in either case we'll have degrees of freedom for our estimates  <= 0.
This also mirrors how both `R` and `scipy` handle such cases.

`scipy.stats.sem` also has a policy for handling `nan` returns, (e.g. throw or return nan) - maybe you could add something similar?

tmcw:
@Yomguithereal @jwilber I've rebased and cleaned up this branch - sorry, wasn't very familiar with standard error the concept, the delay on this one is 😢.

The current state of the branch is that throws with `[]` input, like other methods, and returns 0 on single-element input, which seems like the right approach given that `std_dev / sqrt(length)` returns 0 naively. If that's good for you folks, I'll merge and release with this (thanks @Yomguithereal and sorry for the epic delay :/)

Yomguithereal:
@tmcw so everything's alright with current code? Do you want me to add TS definition file?

src/standard_error.js
```diff
@@ -0,0 +1,24 @@
...
+ * @returns {number} standard error
+ * @example
+ * standardError([2, 4, 4, 4, 5, 5, 7, 9]); // => 2 / Math.sqrt(8)
+ */
+function standardError(x /*: Array<number> */) /*:number*/ {
+    if (x.length === 0) {
+        throw new Error("standardError is not defined for empty arrays");
+    }
+
+    return standardDeviation(x) / Math.sqrt(x.length);
```
dsaxton, line 21: Nitpick: is the name `standardError` too general given that this is only the standard error of the sample mean?

src/standard_error.js
```diff
@@ -0,0 +1,24 @@
...
+ * distribution or sometimes an estimate of that standard deviation.
+ *
+ * @param {Array<number>} x input
+ * @throws {Error} if x has is an empty array
+ * @returns {number} standard error
+ * @example
+ * standardError([2, 4, 4, 4, 5, 5, 7, 9]); // => 2 / Math.sqrt(8)
+ */
+function standardError(x /*: Array<number> */) /*:number*/ {
+    if (x.length === 0) {
```
dsaxton, line 17: Can this be removed and `standardDeviation` itself allowed to throw?
```
> ss.standardDeviation([]) / 0
Uncaught Error: variance requires at least one data point
    at variance (/home/dsaxton/simple-statistics/dist/simple-statistics.js:226:15)
    at Object.standardDeviation (/home/dsaxton/simple-statistics/dist/simple-statistics.js:254:13)
```

MohammedAlkindi:
The 2020 question never got an answer, and `standardError` is still absent from `main`, so nothing here has been superseded.

What moved underneath it since: #807 (2026-07-03) replaced `tap` with `node:test` and shipped the package as ESM, so `test/standard-error.test.js` needs porting and renaming to an underscore, since no hyphenated test files remain. `src/standard_error.js` still carries `/* @flow */` annotations, which `main` no longer has anywhere, and imports `"./standard_deviation"` without the `.js` extension ESM now requires.

A new export today also wants a `src/standard_error.d.ts`, an `index.d.ts` entry, and a changeset. The maths itself is unaffected by all of that.

PR #272, opened 2017-12-07
Update Rollup publishing config

DenisCarriere:
## Update Rollup publishing config

- [x] ~Created symbolic link between `index.js` & `index.d.ts` since they should be the exact same, less chance of forgetting to update the `index.d.ts`~
- [x] ~Add `rollup-plugin-buble` which compiles the code down to ES5 without any complicated configs, in case you want to convert some Simple Statistics modules to pure ES6 (ex: [`bayesian_classifier.js`](https://github.com/simple-statistics/simple-statistics/blob/master/src/bayesian_classifier.js))~
- [x] Add ES format to Rollup, `create-react-app`/`webpack` doesn't play well with bundles that are not pre-compiled.
- [x] Added `browser` to `package.json`, this works for both [Unpkg](https://unpkg.com) & [jsdelivr](https://www.jsdelivr.com)
  - Tested this behavior with `@turf/turf` =>
    - https://cdn.jsdelivr.net/npm/@turf/turf
    - https://unpkg.com/@turf/turf


package.json
```diff
@@ -24,8 +24,9 @@
     "jsdoctest": "^1.7.1",
     "mocha": "^4.0.0",
     "random-js": "^1.0.4",
-    "rollup": "^0.51.0",
-    "rollup-plugin-uglify": "^2.0.1",
+    "rollup": "*",
```
tmcw: I'd rather keep at least a version range here, relying on 'latest' is bound to break in the nearish future?

DenisCarriere: Sounds like a plan!

DenisCarriere: 👍 Updated devDependencies

tmcw:
I'm not sure about the buble move as well - simple-statistics doesn't currently use any ES6 features, and when it does it's unlikely that I'd want a transpilation step - for instance, [modeFast](https://github.com/simple-statistics/simple-statistics/blob/master/src/mode_fast.js) uses an ES6 map for fast equality checks, but I wouldn't want it to ever use a Map polyfill, because that'd negate its speed advantage. BayesianClassifier, for instance - pretty sure it'll work in any ES5 environment?

DenisCarriere:
True... Buble can be removed in your use case since everything is still ES5.

I had to include it in TurfJS since it was getting very hard to handle all the dependencies, since there's 0 dependencies here then it's easy to control.

DenisCarriere:
👍 Removed `buble` from Rollup config

DenisCarriere:
> @tmcw modeFast uses an ES6 map for fast equality checks, but I wouldn't want it to ever use a Map polyfill

From my understanding, `buble` will transpile down to NodeJS 4 (ES2015) compatible, so the `new Map()` should remain unchanged, a lot of the Polyfills won't be added unlike Babel.

Since you aren't use any ES6 features, it's best to leave `buble` out of it since it's not needed.

DenisCarriere:
- [x] Dropped symbolic link for `index.d.ts`, didn't play well when published to `npm`.

@tmcw Only other thing that could be up for discussion is the file format of `"module"`, currently this PR changed it to `dist/simple-statistics.mjs`, but it could easily be changed to `dist/simple-statistics.es.js` if you don't want to force users to use `*.mjs` extension.

Issue #119, opened 2015-08-05
quantile_rank


jonlachlan:
I'm looking for a way to calculate the quantile_rank of a specific data point. For example, I may want to know that a data point is in the 80th percentile, or in the 3rd quartile. I think I'd prefer a precise value, for example "this value is at the 3.245 quartile", which I can round to 3 if I want.

I've done this in Postgres using the ntile() window function (http://www.postgresql.org/docs/9.4/static/functions-window.html), but I'd like to do it in Javascript.

jonlachlan:
I suppose if I have a sorted array, I know the rank of each value, then take that as a percentage of the array length, then multiply by the number of quantiles I'm are looking for.

So if 75 is the 100th element in an array of 200, then this is the 100/200 or .5. Multiply by 100 to get the percentile rank (50) or multiply by 4 for the quartile rank (2).

I think I'm good do this on my own, but thanks anyways :)

tmcw:
so the way to do this would probably be something like

```js
var myNumber = ...;
var myArray = [...];
var qRank = ss.bisect(myArray.sort(), myNumber) / myArray.length;
```

jonlachlan:
Hmm I'm not seeing a bisect function, is that in a new ss dist?

tmcw:
There currently isn't one - I'm proposing we could include one

denizdogan:
Has anyone made any progress on this? I have an unsorted array of values as `[x, y, z, w]` and I'm looking to know what percentile `z` is in, is that a separate issue or is this the one? I noticed that [stats-lite](https://www.npmjs.com/package/stats-lite) has such a function, but since I already use simple-statistics, I'm not too keen on switching now.

Yomguithereal:
@denizdogan it seems the `bisect` function is now part of the current release on npm. Have you tried @tmcw's solution?

Yomguithereal:
What's more, I don't see a function such as the one you need in the `stats-lite` module.

Yomguithereal:
@tmcw on a side note, should we implement a quantile_rank function based on scipy? https://github.com/scipy/scipy/blob/v1.0.0/scipy/stats/stats.py#L1709-L1802

denizdogan:
@Yomguithereal I spoke too quickly, the function in stats-lite is slightly different than the one I proposed. Anyways, I honestly don't understand how I'm supposed to use the `bisect` function, its signature is different than the one in the comment above. :/

[...]

PR #285, opened 2018-03-20
feat: adding quantileRankSorted

Yomguithereal:
Hello everyone, this PR covers the addition of the `quantileRankSorted` function to the library. It's largely inspired by [scipy](https://github.com/scipy/scipy/blob/master/scipy/stats/stats.py) and leverages binary search to work in logarithmic time. I would like your opinion about all that because I am unsure I did everything correctly.

If you feel this everything is ok and valuable to the library, I'll edit the PR to add `quantileRank` also which will very simply copy and sort the array before giving it to `quantileRankSorted`.

Related to #119


src/quantile_rank_sorted.js
```diff
@@ -0,0 +1,93 @@
...
+    return mean / n;
+}
+
+function lowerBound(x, value) {
+    var mid = 0;
+    var lo = 0;
+    var hi = x.length;
+
+    while (lo < hi) {
+        mid = (lo + hi) >>> 1;
```
tmcw: (let me know if my code reading here is wrong)

It looks like this is an optimization of `/2` - and at least in one bench https://jsperf.com/bitwise-vs-multiply-vs-divide-various-number-lengths that looks like an optimization that recent JavaScript engines will do for you.

src/quantile_rank_sorted.js
```diff
@@ -0,0 +1,93 @@
...
+ */
+function quantileRankSorted(
+    x /*: Array<number> */,
+    value /*: number */)/*: number */ {
+
+    var n = x.length;
+
+    var l = lowerBound(x, value);
+
+    // Value is lesser than any value in the array
```
tmcw: From my reading, we can do these early-returns _before_ running `lowerBound`, which for those cases would save a few cycles.

src/quantile_rank_sorted.js
Yomguithereal: It also performs integer division in a single operation though. The alternative would be `Math.trunc((lo + hi) / 2)` or `((lo + hi) / 2) | 0`. Tell me if you prefer one of those two.

src/quantile_rank_sorted.js
Yomguithereal: Correct. I'll update this soon.

src/quantile_rank_sorted.js
```diff
@@ -0,0 +1,93 @@
...
+    if (value < x[0]) {
+        return 0;
+    }
+
+    // Value is greater than any value in the array
+    if (value > x[x.length - 1]) {
```
Yomguithereal: I should probably use `n` here and avoid `x.length` repetition.

Yomguithereal:
@tmcw I have updated the PR.

Yomguithereal:
I also added `quantileRank`.

src/quantile_rank.js
```diff
@@ -0,0 +1,29 @@
...
+ * quantileRank([4, 3, 2, 3, 1], 3); // => 0.7
+ * quantileRank([2, 4, 1, 3], 6); // => 1
+ * quantileRank([5, 3, 1, 2, 3], 4); // => 0.8
+ */
+function quantileRank(
+    x /*: Array<number> */,
+    value /*: number */)/*: number */ {
+
+    // Cloning and sorting the array
+    var sortedCopy = x.slice().sort();
```
tmcw: I suspect there's a bug here: `.sort()` sorts lexicographically, so numbers aren't sorted in order. We'd need to use `numericSort` here.

Yomguithereal: I did not remember this js quirk. Let me update the PR then.

Yomguithereal:
@tmcw changed sorting.

tmcw:
🙏 Released as 5.3.0!

Issue #271, opened 2017-12-05
Nonparametric Tests


jwilber:
Any interest in adding nonparametric tests?
I can start with a permutation test implementation, if so.

tmcw:
Absolutely,  I think that'd be a great addition!

PR #288, opened 2018-04-14
new feature: permutation test

jwilber:
Hello everyone -

Here's an implementation of `permutation_test` (see issue #271 ); hopefully the first of many statistical tests to be added :)

I opted for the `alternative` argument as opposed to `null` because that's congruent with the APIs for similar `R` functions (e.g. `coin::independence_test`, `t.test`, `wilcox.test`), so I figured I'd keep it.

### DIFF
--- a/src/permutation_test.js
+++ b/src/permutation_test.js
@@ -27,13 +27,20 @@ import shuffleInPlace from './shuffle_in_place';
 function permutationTest(
     sampleX/*: Array<number> */,
     sampleY/*: Array<number> */,
-    alternative/*: string */ = 'two_side',
-    k/*: number */ = 10000)/*: ?number */ {
+    alternative/*: string */,
+    k/*: number */)/*: ?number */ {
+    // Set default arguments
+    if (k === undefined) {
+        k = 10000;
+    }
+    if (alternative === undefined) {
+        alternative = 'two_side';
+    }
     if (alternative !== 'two_side' && alternative !== 'greater' && alternative !== 'less') {
-        throw new Error('`alternative` must be either \'two_tail\', \'greater\', or \'less\'');
+        throw new Error('`alternative` must be either \'two_side\', \'greater\', or \'less\'');
     }
 
-    // int pValue var
+    // init pValue
     var pValue;
 
     // get means for each sample
--- a/src/permutation_test.js
+++ b/src/permutation_test.js
@@ -65,23 +65,33 @@ function permutationTest(
         testStatDsn[i] = permTestStatistic;
     }
 
-    // Calculate p-value depending on alternative 
-    // more info on p-value calculations: https://onlinecourses.science.psu.edu/stat464/node/35
+    // Calculate p-value depending on alternative
+    // For this test, we calculate the percentage of 'extreme' test statistics (subject to our hypothesis)
+    // more info on permutation test p-value calculations: https://onlinecourses.science.psu.edu/stat464/node/35
+    var numExtremeTStats = 0;
     if (alternative === 'two_side') {
-        pValue = sum(testStatDsn.map(function(tStat) {
-            return +(Math.abs(tStat) >= Math.abs(testStatistic));
-        })) / k;
+        for (i = 0; i <= k; i++) {
+            if (Math.abs(testStatDsn[i]) >= Math.abs(testStatistic)) {
+                numExtremeTStats += 1;
+            }
+        }
     } else if (alternative === 'greater') {
-        pValue = sum(testStatDsn.map(function(tStat) {
-            return +(tStat >= testStatistic);
-        })) / k;
+        for (i = 0; i <= k; i++) {
+            if (testStatDsn[i] >= testStatistic) {
+                numExtremeTStats += 1;
+            }
+        }
     } else { // alternative === 'less'
-        pValue = sum(testStatDsn.map(function(tStat) {
-            return +(tStat <= testStatistic);
-        })) / k;
+        for (i = 0; i <= k; i++) {
+            if (testStatDsn[i] <= testStatistic) {
+                numExtremeTStats += 1;
+            }
+        }
     }
+
+    pValue = numExtremeTStats / k;
             
-    return pValue
+    return pValue;
         
 }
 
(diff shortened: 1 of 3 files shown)


src/permutation_test.js
```diff
@@ -0,0 +1,85 @@
...
+    // the generated test statistics.
+    var testStatistic = meanX - meanY;
+    
+    // create test-statistic distribution
+    var testStatDsn = new Array(k);
+    
+    for (var i = 0; i < k; i++) {
+        
+        // 1. shuffle data assignments
+        var allData = sampleX.concat(sampleY);
```
Yomguithereal: This should be constructed outside the loop so it's not re-created each time.

src/permutation_test.js
```diff
@@ -0,0 +1,85 @@
...
+        var permData = shuffle(allData);
```
Yomguithereal: You would rather shuffle `allData` in place to avoid allocation unnecessary memory. This can be done because `allData` was constructed by your function and does not belong to the user, so it's fine to mutate it.

src/permutation_test.js
```diff
@@ -0,0 +1,85 @@
...
+        var permLeft = permData.splice(0, Math.floor(permData.length / 2));
```
Yomguithereal: No need to splice and alter array sizes if you compute the mean by yourself below. Tricky thing is we would somehow need a sum function being able to take an offset and a length.

src/permutation_test.js
```diff
@@ -0,0 +1,85 @@
...
+        var permTestStatistic = mean(permLeft) - mean(permRight);
+
+        // 3. store test statistic to build test statistic distribution
+        testStatDsn[i] = permTestStatistic;
+    }
+
+    // Calculate p-value depending on alternative 
+    // more info on p-value calculations: https://onlinecourses.science.psu.edu/stat464/node/35
+    if (alternative === 'two_side') {
+        pValue = sum(testStatDsn.map(function(tStat) {
```
Yomguithereal: Here it would be nice not to have to create a new array to avoid memory allocation but this means rolling your own sum and it could get tedious.

jwilber:
@Yomguithereal excellent advice.

I moved `allData` out of the loop & shuffled in-placed (required changing `splice` to `slice` also).

As for the rolling mean, I think that's a great optimization, but not sure if it's the best for this library, as it may make the code less straightforward (which appears to be a high-priority for the library).

I think the current implementation lends is easily understood when compared to what's going on;
e.g.

<a href="https://www.codecogs.com/eqnedit.php?latex=\text{p-value}_{\text{two-tailed}}=\frac{\text{number&space;of}&space;|D|'s\ge&space;|D_{\text{obs}}|}{{{m&plus;n}\choose&space;m}}" target="_blank"></a>

Let me know what you think!

src/permutation_test.js
mourner: Since here all the parts of the sum are either `0` or `1`, we don't have to use the fancy error-correcting algorithm. It's both much faster and simpler to just use straight `for (...)` loops instead of `sum`/`map`.

src/permutation_test.js
```diff
@@ -0,0 +1,89 @@
...
+    // calculate initial test statistic. This will be our point of comparison with
+    // the generated test statistics.
+    var testStatistic = meanX - meanY;
+
+    // create test-statistic distribution
+    var testStatDsn = new Array(k);
+    
+    // combine datsets so we can easily shuffle later
+    var allData = sampleX.concat(sampleY);
+    var midArray = Math.floor(allData.length / 2);
```
mourner: Let's call that `midIndex` or something like this, since `midArray` implies the variable is an array which it isn't.

src/permutation_test.js
```diff
@@ -0,0 +1,89 @@
...
+    
+    for (var i = 0; i < k; i++) {
+        
+        // 1. shuffle data assignments
+        allData = shuffle(allData);
```
mourner: Instead of allocating a new array here `k` times, you can use [shuffleInPlace](https://github.com/simple-statistics/simple-statistics/blob/master/src/shuffle_in_place.js), since we don't care about `allData` array being shuffled after each iteration or anywhere else in the function.

src/permutation_test.js
```diff
@@ -0,0 +1,98 @@
...
+    
+    // combine datsets so we can easily shuffle later
+    var allData = sampleX.concat(sampleY);
+    var midIndex = Math.floor(allData.length / 2);
+    
+    for (var i = 0; i < k; i++) {
+        
+        // 1. shuffle data assignments
+        shuffleInPlace(allData);
+        var permLeft = allData.slice(0, midIndex);
```
Yomguithereal: Here we are still allocating memory and we could avoid doing it but I thinks it's something to do in the future and for now it's good enough.

mourner: Yeah, thought about this too — to fix this gracefully, we would need to change `mean` to accept optional `left`/`right` index arguments, but then we would need to do that for most of the API surface for consistency. I'm not sure it's worth the increased API complexity.

Yomguithereal: I think it would be nice to have a secondary function, more specialized, exported by the `mean` module such as `meanOffset` or whatever name. But this is an idea for the future :).

src/permutation_test.js
```diff
@@ -0,0 +1,98 @@
...
+ * @example
+ * var control = [2, 5, 3, 6, 7, 2, 5];
+ * var treatment = [20, 5, 13, 12, 7, 2, 2];
+ * permutationTest(control, treatment); // ~0.1324 
+ */
+function permutationTest(
+    sampleX/*: Array<number> */,
+    sampleY/*: Array<number> */,
+    alternative/*: string */ = 'two_side',
+    k/*: number */ = 10000)/*: ?number */ {
```
tmcw: In general this looks great - one thing is that this is using a default parameter, which has more [limited browser support](https://kangax.github.io/compat-table/es6/#test-default_function_parameters) than simple-statistics currently has - this would drop IE11 support, for instance. Mind replacing this with an equivalent like

```js
if (k === undefined) {
  k = 10000;
}
```

tmcw:
😀 Awesome, thanks @jwilber!

PR #290, opened 2018-04-23
Gamma Function

jwilber:
Here's an implementation of the [Gamma Function](https://en.wikipedia.org/wiki/Gamma_function), which is just a continuous approximation to factorial.

I wrote this because I wanted to implement some other methods that require particular distributions, that themselves require the gamma function.

I think most implementations use the Stirling approximation, but I found a paper that claims to have a superior implementation [(Nemes approximation)](https://arxiv.org/abs/1003.6020), so I just used that bad boy.

For negative inputs, I used [Euler's reflection formula](https://en.wikipedia.org/wiki/Gamma_function#Properties) , which I found via [Gyu Eun Lee's answer on math.stachexchange](https://math.stackexchange.com/questions/396889/how-to-find-the-factorial-of-a-fraction).
The results yield identical results to `R`'s `gamma` function, with more decimal points:

```
# r results:
gamma(12.6) # 175523299
gamma(-22.3) # -1.354864e-21
gamma(120.7) # 1.589497e+198

// js results (my implementation):
gamma(12.6)  // 175523299.46913576
gamma(-22.3) // -1.3548635139663688e-21
gamma(120.7) // 1.5894968726397824e+198
```

Also, technically the gamma function is defined for complex inputs as well, but mine doesn't handle such inputs (not yet, anyway). I don't think this is an issue; the `R` version of the function isn't defined for complex inputs either, and I don't think we'll need anything niche/advanced enough so as to require it.

If anyone has suggestions for better code please leave a comment!

### DIFF
--- a/src/gamma.js
+++ b/src/gamma.js
@@ -1,0 +1,58 @@
+/* @flow */
+
+import factorial from './factorial';
+
+/**
+ * Compute the [gamma function](https://en.wikipedia.org/wiki/Gamma_function) of a value using Nemes' approximation.
+ * The gamma of n is equivalent to (n-1)!, but unlike the factorial function, gamma is defined for all real n except zero 
+ * and negative integers (when NaN is returned). Note, the gamma function is also wel-defined for complex numbers, 
+ * though this implementation currently does not handle complex numbers as input values.
+ * Nemes' approximation is defined [here](https://arxiv.org/abs/1003.6020) as Theorem 2.2.
+ * Negative values use [Euler's reflection formula](https://en.wikipedia.org/wiki/Gamma_function#Properties) for computation.
+ *
+ * @param {number} n Any real number except for zero and negative integers.
+ * @returns {number} The gamma of the input value.
+ *
+ * @example
+ * gamma(11.5); // 11899423.084037038
+ * gamma(-11.5); // 2.29575810481609e-8 
+ * gamma(5); // 24 
+ */
+function gamma(n /*: number */)/*: number */ {
+    // Decrement n, because approximation is defined for n - 1
+    n -= 1;
+    
+    if (Number.isInteger(n)) {
+        if (n <= 0) {
+            // gamma not defined for zero or negative integers
+            return NaN;
+        } else {
+            return factorial(n);
+        }
+    }
+    
+    if (n < 0) {
+        // Use Euler's reflection formula for negative inputs
+        // see:  https://en.wikipedia.org/wiki/Gamma_function#Properties
+        return Math.PI / (Math.sin(Math.PI * -n) * gamma(-n));
+        
+    } else {
+        // Nemes' expansion approximation
+        var seriesCoefficient = Math.pow((n / Math.E), n) * Math.sqrt(2 * Math.PI * (n + (1 / 6)));
+
+        var seriesDenom = n + 1 / 4;
+
+        var seriesExpansion = (1 +
+            (1 / 144) / Math.pow(seriesDenom, 2) -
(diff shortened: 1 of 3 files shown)


src/gamma.js
```diff
@@ -0,0 +1,58 @@
...
+ * @returns {number} The gamma of the input value.
+ *
+ * @example
+ * gamma(11.5); // 11899423.084037038
+ * gamma(-11.5); // 2.29575810481609e-8 
+ * gamma(5); // 24 
+ */
+function gamma(n /*: number */)/*: number */ {
+    // Decrement n, because approximation is defined for n - 1
+    n -= 1;
```
tmcw: `n--;` would be a bit more idiomatic for JavaScript.

src/gamma.js
```diff
@@ -0,0 +1,58 @@
+/* @flow */
+
+import factorial from './factorial';
+
+/**
+ * Compute the [gamma function](https://en.wikipedia.org/wiki/Gamma_function) of a value using Nemes' approximation.
+ * The gamma of n is equivalent to (n-1)!, but unlike the factorial function, gamma is defined for all real n except zero 
+ * and negative integers (when NaN is returned). Note, the gamma function is also wel-defined for complex numbers,
```
tmcw: small typo - `wel-defined` should be `well-defined`

test/gamma.test.js
```diff
@@ -0,0 +1,31 @@
...
+    t.test('gamma for positive real number should be correct', function(t) {
+        t.equal(ss.gamma(11.54), 13098426.039156161);
+        t.end();
+    });
+    t.test('gamma for negative real float should be correct', function(t) {
+        t.equal(ss.gamma(-42.5), -3.419793520724856e-52);
+        t.end();
+    });
+    t.test('gamma for negative integer should return NaN', function(t) {
+        t.equal(isNaN(ss.gamma(-2)), true);
```
tmcw: You can write this instead, to save some chars and improve test output:

```
 t.ok(isNaN(ss.gamma(-2)));
```

jwilber:
@tmcw updated - thanks for the style tips!

jwilber:
Just found a quick issue for `gamma(1)` (which should return 1), pushing fix now

jwilber:
Ok, change fixed. 
(if interested, the issue was that decrementing `n` before the integer check caused `gamma` to return `nan` instead of `1`. I rearranged things to yield correct output)

Yomguithereal:
There might be some insights in [this](https://github.com/josdejong/mathjs/blob/d7d25e83c835c3f060bff2e26e293710b61fc32b/lib/function/probability/gamma.js) implementation also (Lanczos, Stirling etc.) but I really don't know about the gamma function implem enough to know if you'll find relevant things there.

src/gamma.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {number} n Any real number except for zero and negative integers.
+ * @returns {number} The gamma of the input value.
+ *
+ * @example
+ * gamma(11.5); // 11899423.084037038
+ * gamma(-11.5); // 2.29575810481609e-8 
+ * gamma(5); // 24 
+ */
+function gamma(n /*: number */)/*: number */ {
+    // Decrement n, because approximation is defined for n - 1
```
tmcw: This comment needs to move along with the line that it describes

jwilber:
@Yomguithereal cool! Looks like we used different approximations, but still yield the same values. Both of ours also hit integer overflow at ~`n=170`.
Theirs can handle complex numbers - perhaps at some point in the future I (or someone else) will pour through the code, so we can adapt complex inputs here. Thanks!

PR #306, opened 2018-05-19
Add datalib

domoritz:


Yomguithereal:
Thanks @domoritz. Would you be up to adding this library's functions to the benchmark also?

Yomguithereal:
On a side note, I see the lib you added is just under vega-statistics. Do you know if this lib is somehow meant to replace it? Because in that case I guess we can drop the earlier link, no?

tmcw:
Thanks @domoritz !

PR #338, opened 2018-10-31
feat: add covariance definition function

NitinSingh2020:
## Description

PR to add [covariance](https://en.wikipedia.org/wiki/Covariance) definition function

### DIFF
--- a/src/covariance.d.ts
+++ b/src/covariance.d.ts
@@ -1,0 +1,6 @@
+declare function covariance(
+    data1: Array<number>,
+    data2: Array<number>
+): number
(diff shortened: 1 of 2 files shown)


tmcw:
Thanks for the PR! As far as I can tell, this is the same as the pre-existing [sample covariance function](https://github.com/simple-statistics/simple-statistics/blob/master/src/sample_covariance.js) - is there something that the existing function is missing?

tmcw:
Closing: I'm pretty sure this is the same as what we already have covariance.

PR #340, opened 2018-11-26
chore(package): update flow-bin to version 0.86.0

tmcw:
Closes #331


tmcw:
I'm going to remove flow instead.

PR #348, opened 2018-12-29
No flo

tmcw:


xixixao:
Heya, hope you don't mind me asking, why was the Flow support removed?

tmcw:
A number of reasons:

- Flow instituted a new degree of caution in some of our method interfaces, but once we switched to that the benefits were very narrow. The class of errors that it reduces (type errors) were always extremely rare in Simple Statistics: this is a pretty compact module with well-defined interfaces and few complex types.
- Having annotations added another hurdle for contributors - we would get PRs without flow annotations and folks would have to learn a subset of the language to add them.
- Flow added serious build & test complexity: we'd have to test transpiled / type-stripped code, or we'd have to use the barely-supported Flow comments syntax.
- The standards around a module 'having flow support' for module users was always vague, ill-defined, and shifting. Simple-statistics already grapples with the convoluted and, well, bad, JavaScript module ecosystem and one more moving target of an entrypoint was not good.
- TypeScript won the popularity contest, so if there were going to be types, then it'd be TypeScript rather than Flow. But again, the overhead of types in terms of maintenance, complexity, and contribution negate the very, very narrow benefits to this kind of project which is much more concerned with logic bugs than type errors.

xixixao:
Thanks @tmcw, this is instructive. Feels like it's really mostly about the last point, given that there are typescript definitions? Or are those potentially not correct?

tmcw:
There are TypeScript definitions, but that's quite a bit different than writing the project and internally annotating functions with typescript, or running the whole thing through the TS compiler. We provide TS definitions for people who want them, and use `tsc` to check their validity, but the Simple Statistics source is in vanilla JavaScript. Previously, the source was in Flow. So it's not really replacing Flow with TS as much as removing Flow and adding separate TS definitions only for module consumers.

Issue #355, opened 2019-01-25
Horner's method


harunorimurata:
`errorFunction()` can be match faster and more accurate using Horner's method.
https://en.wikipedia.org/wiki/Horner%27s_method

Benchmark:
Firefox v64.0.2 64bit, macOS Mojave
```
origin: 204ms
horner: 48ms
```

Code:
```js
function errorFunction(x) {
    const t = 1 / (1 + 0.5 * Math.abs(x));
    const tau =
        t *
        Math.exp(
            -Math.pow(x, 2) -
                1.26551223 +
                1.00002368 * t +
                0.37409196 * Math.pow(t, 2) +
                0.09678418 * Math.pow(t, 3) -
                0.18628806 * Math.pow(t, 4) +
                0.27886807 * Math.pow(t, 5) -
                1.13520398 * Math.pow(t, 6) +
                1.48851587 * Math.pow(t, 7) -
                0.82215223 * Math.pow(t, 8) +
                0.17087277 * Math.pow(t, 9)
        );
    if (x >= 0) {
        return 1 - tau;
    } else {
        return tau - 1;
    }
}

function errorFunction_horner(x) {
    const t = 1 / (1 + 0.5 * Math.abs(x));
    const tau =
        t *
        Math.exp(
            -Math.pow(x, 2)
            - ((((((((0.17087277
                * t - 0.82215223)
                * t + 1.48851587)
                * t - 1.13520398)
                * t + 0.27886807)
                * t - 0.18628806)
                * t + 0.09678418)
                * t + 0.37409196)
                * t + 1.00002368)
                * t - 1.26551223
        );
    if (x >= 0) {
        return 1 - tau;
    } else {
        return tau - 1;
    }
}

var consoleTime = (name, times, func, arg) => {
  console.time(name);
  while (--times) func(arg);
  console.timeEnd(name);
}

const x = Math.random();
consoleTime('origin', 100000, errorFunction, x);
consoleTime('horner', 100000, errorFunction_horner, x);
```

tmcw:
Looks good to me - I do wish that our implementation explained a bit more where these magic numbers are coming from (0.17087277) because afaict they aren't something explained by the linked Wikipedia pages.

harunorimurata:
I couldn't find a good online document about the magic numbers other than the Wikipedia page...

tmcw:
Implemented in https://github.com/simple-statistics/simple-statistics/pull/356!

PR #356, opened 2019-01-26
Refactor: Apply Horner's method to error function

harunorimurata:
Applying Horner's method to `errorFunction`.
Related to #355.


src/error_function.js
```diff
@@ -16,17 +16,21 @@ function errorFunction(x) {
     const tau =
         t *
         Math.exp(
-            -Math.pow(x, 2) -
```
tmcw: Mind adding more information to the JSDoc comment above this method to explain which algorithm is used to solve it?

harunorimurata: I just added. Thanks!

PR #357, opened 2019-01-31
offload the bulk of the work of min/max to Math.max/min

a-hariti:
I know that the goal of the project is to have straight forward code, but I think that using `Math.max/min` wouldn't hinder that, while being ~~faster~~.

### DIFF
--- a/src/max.js
+++ b/src/max.js
@@ -15,15 +15,7 @@ function max(x) {
         throw new Error("max requires at least one data point");
     }
 
-    let value = x[0];
-    for (let i = 1; i < x.length; i++) {
-        // On the first iteration of this loop, max is
-        // undefined and is thus made the maximum element in the array
-        if (x[i] > value) {
-            value = x[i];
-        }
-    }
-    return value;
+    return Math.max.apply(null, x);
 }
 
 export default max;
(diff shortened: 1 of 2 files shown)


Yomguithereal:
It might depend on the size of the list, `Math.min/max` being variadic and the processing of arguments being quite slow on some engines. Can you try a minor benchmark to check this please? For instance this jsperf bench [here](https://jsperf.com/min-and-max-in-array/3) shows that using `Math` is the worst solution of all. (On my computer, reduce wins buts this is probably because the bench's array is ridiculously small).

a-hariti:
well, I'm not a big fan of micro-benchmarks because one can never be sure, for example `Array.reduce` suffers greatly  once you increase the size of the input array (1000s of elements).
and even then one still can not draw any solid conclusions, for example testing with diffrent input arrays is not the same as testing with one (caching ...)

I think that in order to interpret such results, one needs to outsmart the js engine (I can't) 
ps: the  `loop` version _seems_ to **win** every time.

tmcw:
Thanks for the contribution! As Guillaume points out, Math.max/min are slower than the current solution, and there's an additional downside: there's a function argument length limit. For example:

```js
Math.max(...Array.from({ length: 500000 }, (_, i) => i))
```

Will fail on Firefox, because the limit is 500,000 arguments.

PR #380, opened 2019-06-25
chore(package): update tap to version 14.3.1

tmcw:


tmcw:
This, and other test failures, are rooted in https://github.com/vfile/vfile-message/issues/3

PR #381, opened 2019-07-06
export permutationTest for ts

alope107:
Previously `index.d.ts` did not export `permutationTest`. This fixes that.


tmcw:
Thanks! Rebased and merged in #383

PR #430, opened 2020-01-22
add missing typings for extent

chriszrc:
It appears the typings for extent are not currently being exported ( extent itself is), this adds the types file for extent, and adds it to the index.d.ts exports


src/extent.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * https://simplestatistics.org/docs/#max
```
tmcw: typo - max here should be extent

chriszrc: Ah, good catch, yes indeed, ok, fixed-

PR #449, opened 2020-04-20
build: Node v13 import support

tmcw:
Node 13 includes another way to include modules. This is terrible, the only way to explain a modern package.json is with excuses.


tmcw:
You have to set type: module, which turns off the ability to use CommonJS. What a trash fire.

PR #455, opened 2020-05-05
Simple mean

flagpoonage:
Internally, the standard `mean` function is using a `sum` function which goes beyond successive addition to correct for floating-point errors. The benchmarked `mean` function in the jStat library does not provide this, so the benchmark is not testing against which code is more performant, but instead testing two different algorithms.

Since the `sumSimple` function is already provided, it follows on that `meanSimple` can be used to provide an end user with an option for a more performant mean, at the cost of precision. This also allows the benchmark against jStat to be far closer, and in my tests actually outperforming.


tmcw:
Awesome - good idea, thanks! Merged & released as 7.1.0.

PR #483, opened 2020-08-16
Silhouette measure of cluster fit

gvwilson:
1.  Implement the [silhouette method](https://en.wikipedia.org/wiki/Silhouette_(clustering)) for evaluating goodness-of-fit of clusters with both vector and scalar results.
2.  Create the first few tests.
3.  Move `makeMatrix` into its own file so that it can be re-used during silhouette calculations.
4.  Put Euclidean distance measure in its own file so that it can be shared between k-means clustering and silhouette measure.

To do:

1.  More tests.
2.  TypeScript support files - guidance on these would be welcome.
3.  More documentation: should every top-level function include at least one `@example`? And is there a convention better than `Array<Array<number>>` to indicate rectangular arrays-of-arrays?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -92,6 +92,9 @@ export { default as permutationTest } from "./src/permutation_test";
 // Root-finding methods
 export { default as bisect } from './src/bisect';
 
+// Clustering methods
+export { default as kMeansCluster } from './src/k_means_cluster';
+
 // Utils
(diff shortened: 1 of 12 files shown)


tmcw:
Sorry for the delay - I'll take a look and get a PR later tonight or tomorrow!

PR #489, opened 2020-08-27
Fixing documentation link

gvwilson:
`GETTTING` with three T's should be `GETTING`.

Also, I cannot find a `documentation.yml` file in this repository - where and how are section headings in the documentation in https://simplestatistics.org/docs/ determined?


tmcw:
Looks good, but please remove the package-lock - there's a yarn.lock and this project uses yarn. The documentation.yml file is here: https://github.com/simple-statistics/simple-statistics.github.com/blob/master/docs/documentation.yml

PR #492, opened 2020-08-28
Making relative error and approximate equality available to users.

### LINKED ISSUE #486: Make testing utilities in test/approx.js available as part of the package
1.  People writing applications that use this library need to test their programs.
2.  The utility functions in `test/approx.js` are useful for doing this.
3.  Move them to `src/` and make them generally available.
4.  Provide a generalized comparator to implement "approximately deepEqual".

If this proposal is accepted, I volunteer to do the work.

gvwilson:
1.  Move relative error and shallow/deep approximate equality out of `test/approx.js` helper file and into `src/relative_error.js`, `src/approx_strict_equal.js`, and `src/deep_approx_strict_equal.js`.
1.  Document.
1.  Add `.d.ts` files.
1.  Add entries to `./index.js`.
1.  Add tests.
1.  Modify `test/silhouette.test.js` to use new functions.

Closes #486.


src/deep_approx_strict_equal.js
```diff
@@ -0,0 +1,118 @@
...
+    // Booleans must be exactly equal
+    else if (typeof expected === "boolean") {
+        return expected === actual;
+    }
+
+    // Strings must be exactly equal
+    else if (typeof expected === "string") {
+        return expected === actual;
+    }
+
```
tmcw, line 44: I think that the expansion to other types makes the meaning of this function a little weaker: this is only approximate for numbers, but strict for everything else. Especially if we enforce a strict TypeScript definition (only numbers and maybe arrays of numbers), and strict documentation, there's little reason to worry about people accidentally using this method for strings or any kind of unknown datatype: they're using a custom approximate equal method on purpose, and if they want another type of equality, they have implementations for that close at hand with the  `assert` module or the `===` operator.

src/deep_approx_strict_equal.js
```diff
@@ -0,0 +1,118 @@
...
+        return arrayApproxStrictEqual(actual, expected, tolerance);
+    }
+    else if (Array.isArray(actual)) {
+        return false;
+    }
+
+    // Objects must have matching keys and approximately equal values
+    else if (typeof expected === "object") {
+        return objectApproxStrictEqual(actual, expected, tolerance);
+    }
```
tmcw, line 61: I'm not sure about the utility of an object implementation here - it only really applies for objects that have numeric values and no prototype differences. Keeping this restricted to numbers and arrays of numbers seems to me like the most straightforward way to implement & document it, without getting into the rather sophisticated ways in which objects can be unequal.

gvwilson: I had two use cases in mind:

1. A function returns multiple values in an object (e.g., `kMeansCluster`). However, this is rare and the parts can be checked for approximate equality one by one.

2. Complex numbers. Libraries like `complex.js` represent these as two-part objects; checking an array of such pairs without an object-handling case can be done with a couple of `map` calls to extract real and imaginary parts, but that feels clumsier than the top-level object decomposition in #1. On the other hand, complex values are much less common in statistics than in signal processing, so this could well be over-engineering.

Conclusion: strip out object handling, only do arrays, add objects back in later if people want them?

tmcw: > Conclusion: strip out object handling, only do arrays, add objects back in later if people want them?

👍 yep, sounds like the right approach. The kMeansCluster makes sense but is pretty rare, and I'm not sure that relative error on the components of a complex number is what folks desire/expect in that case.

tmcw:
Merged with #506 - I did some refactoring and realized that the `deep` case we have is solvable with `Array.every`, so I cut it for this pass: approxEqual is really the key improvement that we want and we can add more later if necessary. Thanks!

PR #514, opened 2020-10-24
Fix "Sample mean and covariance" reference

tniessen:
This reference is either outdated or never worked in the first place.


tmcw (approved):
👍 thanks, must've been an errant rename at some point.

PR #515, opened 2020-10-24
Avoid unnecessary copy in shuffle

tniessen:
The previous implementation of shuffle created two shallow copies, one of which was unnecessary.


tmcw:
Thanks!

PR #518, opened 2020-10-25
Improve use of Big-O notation in src/docs

tniessen:
I believe that this is more readable and closer to being grammatically correct.


tmcw (approved):
👍 Thanks!

Issue #526, opened 2020-11-04
Do we have "Coefficient of Variation" function ?


angelyaaaaa:
I've searched within all document and code, it seems there is no function for Coefficient of Variation 
: https://en.wikipedia.org/wiki/Coefficient_of_variation

I assume it will look like this : https://simplestatistics.readthedocs.io/en/latest/index.html?highlight=coefficient%20variation#coefficient-of-variation

I wonder if there is any function I missed ? or this it's in the plan ?

tmcw:
This function isn't currently included. I'm open to including it - it would likely be a one-line method, given that the [python implementation is so tiny](https://simplestatistics.readthedocs.io/en/latest/_modules/simplestatistics/statistics/coefficient_of_variation.html#coefficient_of_variation) in the meantime, if you want to calculate it, it is just standard deviation divided by the mean.

angelyaaaaa:
cool, I'll send a pr

tmcw:
Thanks - merged and released in 7.4.0!

PR #528, opened 2020-11-23
feat: add coefficient of variation

angelyaaaaa:
#Add coefficient of variation
related to https://github.com/simple-statistics/simple-statistics/issues/526 
- add new function
- follow the formula, using `sample standard deviation` instead of `standard deviation`


angelyaaaaa:
@tmcw could you help review it ? I think it quite simple.

tmcw:
Cleaned up and merged with #535!

PR #549, opened 2021-02-12
Add TypeScript bindings for silhouette.

altearius:
I found that the `silhouette` function was missing TypeScript bindings and could not be imported without them.  The goal of this PR is to add the necessary bindings for this function.


tmcw:
Thanks!

PR #552, opened 2021-02-21
Add logistic distribution utilities

dsaxton:
I think it could be handy to include some functions for dealing with the logistic distribution given its connection with logistic regression. I tried to pick names that would be consistent with the analogous pair of functions `ss.probit` and `ss.cumulativeStdNormalProbability`.


tmcw:
👍 Thanks! I added an additional little test assertion in #553 and merged this in.

PR #554, opened 2021-02-25
feat: Add rank correlation

dsaxton:
This adds a calculation for [Spearman's rank correlation](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient) which only requires us to extract the ranks of the two arrays before calling `sampleCorrelation`. Only caveat is this will use the minimum rank in the case of ties (e.g., `[1, 5, 5, 2, 4, 8]` will become `[0, 3, 3, 1, 2, 5]`).


dsaxton:
Closing in favor of https://github.com/simple-statistics/simple-statistics/pull/562

Issue #171, opened 2017-01-24
Faulty Geometric Mean Algorithm 


BDiehr:
Simple Statistics uses the naive algorithm, which takes the entire product as the first step. This is problematic because large datasets will very often either zero out or hit `Infinity` resulting in an incorrect result.

Similar issue as is discussed in this StackOverflow http://stackoverflow.com/questions/19980319/efficient-way-to-compute-geometric-mean-of-many-numbers

tmcw:
From what I can see, [scipy](https://github.com/scipy/scipy/blob/master/scipy/stats/stats.py#L261-L315) uses the same algorithm, as does [atoll.js](https://github.com/nsfmc/atoll.js/blob/master/lib/atoll.js#L131). Are there any examples of the alternative algorithm in a mainstream statistics library? I'm happy to include an alternative implementation, probably under a different function name (similar to how we have both kahan and naive sums), but would want to confirm that the solution alluded to on stackoverflow (a) is used in practice and (b) has a name or a citation.

BDiehr:
The wikipedia article on the Geometric Mean has a section dedicated to the use of using Log in this manner: https://en.wikipedia.org/wiki/Geometric_mean#Relationship_with_logarithms

> The log form of the geometric mean is generally the preferred alternative for implementation in computer languages because calculating the product of many numbers can lead to an arithmetic overflow or arithmetic underflow. This is less likely to occur with the sum of the logarithms for each number.

They refer to it as the log-average.

I don't have anything beyond that, unfortunately Wikipedia doesn't provide a citation for that passage.

PR #556, opened 2021-03-01
feat: Implement log average

dsaxton:
Closes https://github.com/simple-statistics/simple-statistics/issues/171

Adds a `logAverage` function as an equivalent way of calculating the geometric mean. Also fixes what's arguably a bug in `geometricMean` where it throws an error for inputs containing zero instead of returning zero. Interestingly `logAverage` still handles that case correctly because `Math.log(0)` becomes `-Infinity`, and `Math.exp(-Infinity)` turns out to be zero.


test/log_average.test.js
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    t.test("cannot calculate for lists with negative numbers", function (t) {
+        t.throws(function () {
+            ss.logAverage([-1]);
+        });
+        t.end();
+    });
+
+    t.test("does not overflow for large products", function (t) {
```
dsaxton, line 21: We can also remove this and the underflow test as they're a tad on the trivial side for this function (would be more interesting for `geometricMean`)

PR #561, opened 2021-03-02
Bump prettier and fix lint errors

dsaxton:
This bumps prettier in `yarn.lock` and fixes the resulting lint errors. (I noticed there was a PR from dependabot only bumping the version; not totally sure how dependabot works but I think this PR can supersede that one.)

With this change cloning locally and then simply doing
```
npm install
npm run test
```
should result in clean output.


tmcw (commented):
👍 Thanks!

PR #562, opened 2021-03-02
Rank correlation

tmcw:
@dsaxton for your consideration, decided to take a shot at seeing whether we could support duplicate indexes in this algorithm. This does, by first recording the indexes before sorting.

The benefits are that this supports duplicate values and also improves performance (in big-o terms) because it removes the `indexOf`, which was in the worst case was n².

On the downside, it's a bit more code, and it will use a little more memory, in the form of temporary arrays.


dsaxton:
Nice @tmcw :+1: sounds like this might be a better implementation. I'll go ahead and close the other PR if you'd like to merge this one in.

Issue #395, opened 2019-11-11
Standard normal table docs


bennlich:
From the docs for https://simplestatistics.org/docs/#standardnormaltable:

> The table used is the cumulative, and not cumulative from 0 to mean.

I'm having trouble understanding this sentence. What is meant by cumulative *and* not cumulative?

PR #563, opened 2021-03-03
docs: Update normal CDF doc

dsaxton:
Closes https://github.com/simple-statistics/simple-statistics/issues/395

I think I agree with the original post that this is a little confusing. Maybe just removing the last sentence and moving the link into the paragraph above would help? What do you think @tmcw?


tmcw (approved):
👍 yep, after the refactor that comment didn't make sense. This is much better.

Issue #519, opened 2020-10-26
Docs link is broken


NickHeiner:
The [docs for `tTestTwoSample`](https://simplestatistics.org/docs/#ttesttwosample) link to http://www.monarchlab.org/Lab/Research/Stats/2SampleT.aspx, which 404s.

PR #564, opened 2021-03-03
docs: Modify t-test doc text and link

dsaxton:
Closes https://github.com/simple-statistics/simple-statistics/issues/519

This replaces the existing link with a live one, and makes a small tweak to the wording (philosophical point, but we might want to avoid implying we accept a null whenever we don't reject it).


tmcw (approved):
👍 LGTM!

Issue #250, opened 2017-10-02
Sample performance


Yomguithereal:
Currently, the `sample` function runs in `O(k)` (k being sample size) time but in `O(n)` (n being the size of the population). This can be awful for algorithms sampling large lists frequently (k-means, NN-descent etc...).

I developed some more efficients methods (having their drawback, however) in the [pandemonium](https://github.com/Yomguithereal/pandemonium) library (reservoir sampling still lacking but will soon be added). Does some of the library's method interest `simple-statistics`?

tmcw:
Absolutely! Simple-statistics doesn't have much flexibility in sampling: I started with the Fisher-Yates approach because I could be comfortable with it being suitably random. Would love to add other methods, especially ones that are similarly random-enough and ones that work with streaming data.

PR #565, opened 2021-03-04
Shuffle less during sample

dsaxton:
xref https://github.com/simple-statistics/simple-statistics/issues/250

Made an attempt at improving the efficiency of `sample` a little bit. To do that we're generalizing the `shuffleInPlace` function to `shuffleSubsetInPlace` (open to suggestions for other names; in any case this wouldn't be exposed through the public API) which accepts the number of iterations in the Fisher-Yates shuffle as an additional argument. We then only swap that number of elements and return the selected values from the tail of the array. We can also now treat `shuffleInPlace` as a special case where the number of iterations equals the sample size.

I'm not sure what's going on with the tests, but something with the random number generator seems to be causing output to change. What's being returned now still appears "correct" to me, so I think it should be okay to simply update these.

### DIFF
--- a/src/shuffle_subset_in_place.js
+++ b/src/shuffle_subset_in_place.js
@@ -1,0 +1,35 @@
+function shuffleSubsetInPlace(x, n, randomSource) {
+    // a custom random number source can be provided if you want to use
+    // a fixed seed or another random number generator, like
+    // [random-js](https://www.npmjs.org/package/random-js)
+    randomSource = randomSource || Math.random;
+
+    // store the current length of the x to determine
+    // when no elements remain to shuffle.
+    let length = x.length;
+
+    // temporary is used to hold an item when it is being
+    // swapped between indices.
+    let temporary;
+
+    // The index to swap at each stage.
+    let index;
+
+    // While there are still items to shuffle
+    while (length > Math.max(x.length - n, 0)) {
+        // choose a random index within the subset of the array
+        // that is not yet shuffled
+        index = Math.floor(randomSource() * length--);
+
+        // store the value that we'll move temporarily
+        temporary = x[length];
+
+        // swap the value at `x[length]` with `x[index]`
+        x[length] = x[index];
+        x[index] = temporary;
+    }
+
+    return x.slice(x.length - n);
+}
+
+export default shuffleSubsetInPlace;
--- a/src/sample.js
+++ b/src/sample.js
@@ -1,4 +1,4 @@
-import shuffle from "./shuffle";
+import shuffleSubsetInPlace from "./shuffle_subset_in_place";
 
 /**
  * Create a [simple random sample](http://en.wikipedia.org/wiki/Simple_random_sample)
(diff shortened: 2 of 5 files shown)


src/shuffle_subset_in_place.js
```diff
@@ -0,0 +1,35 @@
...
+
+    // temporary is used to hold an item when it is being
+    // swapped between indices.
+    let temporary;
+
+    // The index to swap at each stage.
+    let index;
+
+    // While there are still items to shuffle
+    while (length > Math.max(x.length - n, 0)) {
```
dsaxton, line 19: This and L32 are really the only ones changing

Yomguithereal:
Hello @dsaxton, I think Fisher-Yates is not the most performant way to produce a sample because it has quite a memory and time cost. I usually use "Algorithm L" to do the same thing in roughly `O(k)` time and space for random access memory. It is described in this paper ([wikipedia](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm)):

> Li, Kim-Hung. "Reservoir-sampling algorithms of time complexity O(n (1+ log (N/n)))." ACM Transactions on Mathematical Software (TOMS) 20.4 (1994): 481-493.

and implemented in JavaScript in [this](https://github.com/Yomguithereal/pandemonium) library of mine. The code is fairly simple and can be found [here](https://github.com/Yomguithereal/pandemonium/blob/master/geometric-reservoir-sample.js).

I am totally fine if `simple-statistics` wants to replicate it to improve the performance of is sample function. Although I would maybe vouch for some kind of heavy testing before to be sure my code is robust and that the produced samples have the necessary qualities.

Note also that as per [this](https://github.com/Yomguithereal/pandemonium#summary) summary, there are a lot of ways to sample an array and for cases where the array is small, Fisher-Yates might be faster and for cases when k << n, the "naive" sample might dominate.

What do you think @tmcw?

dsaxton:
> Hello @dsaxton, I think Fisher-Yates is not the most performant way to produce a sample because it has quite a memory and time cost.

@Yomguithereal Agreed this is not the most efficient algorithm (I'll update the original post to not auto-close the issue); this PR was really just addressing the low-hanging fruit of reducing the time complexity from n to k (but still it copies data unnecessarily). If @tmcw is good with it you may want to submit a pull request which could either replace the implementation here, or create a separate function (then users have the choice if there are cases when one might be preferred over the other).

Yomguithereal:
> Agreed this is not the most efficient algorithm (I'll update the original post to not auto-close the issue); this PR was really just addressing the low-hanging fruit of reducing the time complexity from n to k (but still it copies data unnecessarily).

I agree that this improves the overall performance of the function but since copying is `O(n)`, the time complexity does not change much (it's only decreased by `n - k` operations, which are indeed more costly than the copying one, since those happen on the optimized engine side).

Anyway I also agree that changing the algorithm used by the sampling function should probably happen subsequently to merging this PR, not to mix concerns.

dsaxton:
I'm going to close this since I've convinced myself the better approach would be to just replace the algorithm with a faster implementation as @Yomguithereal has pointed out. I also realized this change is actually incorrect: if we want to take a sample of size one from a data set of size greater than one then that sample has zero probability of being the first element (as a simple counterexample).

PR #567, opened 2021-03-15
build: Expand prettier linting

dsaxton:
@tmcw I noticed `prettier` was only being run against TypeScript declaration files, so I thought it might be helpful to expand it to most of the code base (here including all of `src` and `test` other than the autogenerated `test/types.js` file)


tmcw (approved):
👍 thanks! Not sure why that wasn't already the case, must've been a long-lived accident.

PR #568, opened 2021-03-16
feat: Implement Wilcoxon rank sum statistic

dsaxton:
@tmcw I'm not sure if this is something you'd be interested in adding, but this PR implements a calculation of the Wilcoxon rank sum statistic, which is a popular non-parametric alternative to the two sample t-test. The code gets a tad messy, primarily because we need to make sure that ranks of tied values get replaced by the average rank within the group (else the statistic isn't well-defined). Please let me know if you think there's a more elegant way to do this.

Also worth pointing out that this _only_ calculates the value of the test statistic, and doesn't provide any utilities for performing inference on that value.

### DIFF
--- a/src/wilcoxon_rank_sum.js
+++ b/src/wilcoxon_rank_sum.js
@@ -24,8 +24,11 @@ function wilcoxonRankSum(sampleX, sampleY) {
     const pooledSamples = sampleX
         .map((x) => ({ label: "x", value: x }))
         .concat(sampleY.map((y) => ({ label: "y", value: y })))
-        .sort((a, b) => a.value - b.value)
-        .map((s, idx) => ({ label: s.label, value: s.value, rank: idx }));
+        .sort((a, b) => a.value - b.value);
+
+    for (let rank = 0; rank < pooledSamples.length; rank++) {
+        pooledSamples[rank].rank = rank;
+    }
 
     let tiedRanks = [pooledSamples[0].rank];
     for (let i = 1; i < pooledSamples.length; i++) {
--- a/src/wilcoxon_rank_sum.js
+++ b/src/wilcoxon_rank_sum.js
@@ -1,0 +1,57 @@
+/**
+ * This function calculates the Wilcoxon rank sum statistic for the first sample
+ * with respect to the second. The Wilcoxon rank sum test is a non-parametric
+ * alternative to the t-test which is equivalent to the
+ * [Mann-Whitney U test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test).
+ * The statistic is calculated by pooling all the observations together, ranking them,
+ * and then summing the ranks associated with one of the samples. If this rank sum is
+ * sufficiently large or small we reject the hypothesis that the two samples come
+ * from the same distribution in favor of the alternative that one is shifted with
+ * respect to the other.
+ *
+ * @param {Array<number>} sampleX a sample as an array of numbers
+ * @param {Array<number>} sampleY a sample as an array of numbers
+ * @returns {number} rank sum for sampleX
+ *
+ * @example
+ * wilcoxonRankSum([1, 4, 8], [9, 12, 15]); // => 6
+ */
+function wilcoxonRankSum(sampleX, sampleY) {
+    if (!sampleX.length || !sampleY.length) {
+        throw new Error("Neither sample can be empty");
+    }
+
+    const pooledSamples = sampleX
+        .map((x) => ({ label: "x", value: x }))
+        .concat(sampleY.map((y) => ({ label: "y", value: y })))
+        .sort((a, b) => a.value - b.value)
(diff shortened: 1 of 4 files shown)


src/wilcoxon_rank_sum.js
```diff
@@ -0,0 +1,57 @@
...
+            }
+        } else if (tiedRanks.length > 1) {
+            replaceRanksInPlace(pooledSamples, tiedRanks);
+        } else {
+            tiedRanks = [pooledSamples[i].rank];
+        }
+    }
+
+    function replaceRanksInPlace(pooledSamples, tiedRanks) {
+        const average = (tiedRanks[0] + tiedRanks[tiedRanks.length - 1]) / 2;
```
dsaxton, line 45: Here we can use the structure of `tiedRanks`, since by construction it'll always consist of adjacent integer values in order

src/wilcoxon_rank_sum.js
```diff
@@ -0,0 +1,57 @@
...
+        for (let i = 0; i < tiedRanks.length; i++) {
+            pooledSamples[tiedRanks[i]].rank = average;
+        }
+    }
+
+    return pooledSamples
+        .filter((s) => s.label === "x")
+        .map((s) => s.rank + 1)
```
dsaxton, line 53: Need to add one to go from 0-based indexes to ranks

src/wilcoxon_rank_sum.js
```diff
@@ -0,0 +1,57 @@
...
+function wilcoxonRankSum(sampleX, sampleY) {
+    if (!sampleX.length || !sampleY.length) {
+        throw new Error("Neither sample can be empty");
+    }
+
+    const pooledSamples = sampleX
+        .map((x) => ({ label: "x", value: x }))
+        .concat(sampleY.map((y) => ({ label: "y", value: y })))
+        .sort((a, b) => a.value - b.value)
+        .map((s, idx) => ({ label: s.label, value: s.value, rank: idx }));
```
dsaxton, line 28: `rank` now gives us the correct ranks minus one if we have no duplicates, need to adjust for that next

src/wilcoxon_rank_sum.js
```diff
@@ -0,0 +1,57 @@
...
+
+    let tiedRanks = [pooledSamples[0].rank];
+    for (let i = 1; i < pooledSamples.length; i++) {
+        if (pooledSamples[i].value === pooledSamples[i - 1].value) {
+            tiedRanks.push(pooledSamples[i].rank);
+            if (i === pooledSamples.length - 1) {
```
dsaxton, line 34: Have to special case the "end of the array" since we can't rely on the next value being different from the current to trigger the replacing of tied ranks

src/wilcoxon_rank_sum.js
tmcw, line 28:
```suggestion
    const pooledSamples = sampleX
        .map((x) => ({ label: "x", value: x }))
        .concat(sampleY.map((y) => ({ label: "y", value: y })))
        .sort((a, b) => a.value - b.value);

    for (let rank = 0; rank < pooledSamples.length; rank++) {
        pooledSamples[rank].rank = rank;
    }
```

Not a gigantic change, but since we're doing a lot of o(n) copies, maps, and object constructions, this will save the recreation of an array and the objects in it.

src/wilcoxon_rank_sum.js
```diff
@@ -0,0 +1,57 @@
...
+        const average = (tiedRanks[0] + tiedRanks[tiedRanks.length - 1]) / 2;
+        for (let i = 0; i < tiedRanks.length; i++) {
+            pooledSamples[tiedRanks[i]].rank = average;
+        }
+    }
+
+    return pooledSamples
+        .filter((s) => s.label === "x")
+        .map((s) => s.rank + 1)
+        .reduce((a, b) => a + b);
```
tmcw, line 54:
```suggestion
    let rankSum = 0;

    for (let i = 0; i < pooledSamples.length; i++) {
        const sample = pooledSamples[i];
        if (sample.label === "x") {
            rankSum += sample.rank + 1;
        }
    }

    return rankSum;
```

Similarly, we can save 2 array allocations here and make the code (imho) a bit simpler by using the imperative style.

tmcw (approved):
I added a few suggestions - related to functional vs. imperative style - I think in this case, it makes sense to use imperative style where it'll improve performance and doesn't create any problems with mutability or readability. Besides that, looks great!

dsaxton:
> I added a few suggestions - related to functional vs. imperative style - I think in this case, it makes sense to use imperative style where it'll improve performance and doesn't create any problems with mutability or readability. Besides that, looks great!

Good suggestions, thanks! I tend to write in a functional way out of habit, but these edits do seem more efficient.

dsaxton:
@tmcw gentle ping, this good to merge you think?

tmcw (approved):
👍 Sorry for the delay, looks good!

PR #589, opened 2021-06-24
Add missing TypeScript typing for `silhouetteMetric`

bmish:
The typings were `silhouette` were also forgot in the original implementation (https://github.com/simple-statistics/simple-statistics/pull/485) but that was fixed here: https://github.com/simple-statistics/simple-statistics/pull/549


tmcw (approved):
Thanks!

PR #662, opened 2022-08-12
Support TypeScript nodenext module resolution

abrenneke:
Quick fix, the export map needs a `types` field with nodenext resolution and doesn't use the root `types` field


tmcw (approved):
Great, thanks!

PR #673, opened 2022-10-26
Bring back jenks

### LINKED ISSUE #672: Bring back Jenks
Even though Ckmeans may be superior to Jenks, the Jenks algorithm is of historical interest. It would be great to have a working implementation of Jenks around for reference and use. For me personally, I'm working on a book about data visualization and I'd like to include an example that uses Jenks. The deprecated implementation in this library seems to be the best around.

FWIW I landed here after reading [Literate Jenks Natural Breaks and How The Idea Of Code is Lost](https://macwright.com/2013/02/18/literate-jenks.html) and searching for the source.

It is possible to try Jenks now by using an older build of this library, 0.9.2, but would it be a possibility to bring back Jenks to the official current API? I think the implementation of this change would be fairly trivial, just bringing back the [old files that were removed](https://github.com/simple-statistics/simple-statistics/pull/118/files#diff-b95ab92f03e4150ced5c6fcaa9b82d607aa74a04de21f6b07a12f10a24f56d50)

I'd love to ultimately make something like [ckmeans & now deprecated Jenks Natural Breaks with simple-statistics and d3](http://bl.ocks.org/thadk/37ff4c03358b414acd1e), but without hackery around using older versions.

Is this something you'd be open to? Thanks!

curran:
Closes #672

I brought back the original files, then modified them to adhere to the latest code style (using ES modules and let/const), then made sure linting and tests pass.

The only thing missing is the TS declarations, which I'm not sure how to do properly.

I hope this is a welcome change!

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,6 +37,7 @@ export { default as shuffle } from "./src/shuffle.js";
 export { default as shuffleInPlace } from "./src/shuffle_in_place.js";
 export { default as sample } from "./src/sample.js";
 export { default as ckmeans } from "./src/ckmeans.js";
+export { default as jenks } from "./src/jenks.js";
 export { default as uniqueCountSorted } from "./src/unique_count_sorted.js";
 export { default as sumNthPowerDeviations } from "./src/sum_nth_power_deviations.js";
 export { default as equalIntervalBreaks } from "./src/equal_interval_breaks.js";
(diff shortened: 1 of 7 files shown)


tmcw:
Merged with a cleanup branch in #674, thanks!

Issue #704, opened 2023-03-13
Unexpected results for multiple quantiles


donmccurdy:
I'm seeing an issue very similar to #262 when using the latest release (v7.8.3). These methods give very different results for the p05 quantile —

```javascript
const a = quantile(data, [0.0, 0.05, 0.5, 0.95, 1.0]);
const b = quantile(data, [0.0, 0.05,      0.95, 1.0]);
const c = [
  quantile(data, 0.0),
  quantile(data, 0.05),
  quantile(data, 0.95),
  quantile(data, 1.0)
];
```

For my input data, A and C match, but B[1] is incorrect. 

| case | p00 |	p05 | p50 | p95 | p100 |
|------|-----|-----|-----|-----|------|
| A | 0 | 3,569,032 | 8,370,252 | 11,183,760 | 17,198,303 |
| B | 0 | 8,370,252 ❌ | — | 11,183,760 | 17,198,303 |
| C | 0 | 3,569,032 | — | 11,183,760 | 17,198,303 |

<details>
<summary>input.json</summary>

```json
[
    4534264.499853279,
    0,
    0,
    4441655.968117084,
    7982031.499157332,
    11183760.330270268,
    8374531.7466523405,
    7537166.851276029,
    6408451.109897822,
    4858231.646338846,
    6409803.519554875,
    4296216.572964368,
    5367514.3078208575,
    6364913.420875475,
    8621150.51996475,
    9236261.500350267,
    7850954.174512796,
    6026897.449805159,
    3569031.9331034794,
    5336662.162307054,
    6953168.388225255,
    6789669.207787706,
    6225885.17604323,
    5217135.654872652,
    6199349.484675659,
    5858452.126622974,
    7235692.6957654515,
    5129514.715977198,
    5161771.078309817,
    6488305.210989551,
    8371033.429666803,
    6753743.962686865,
    6595808.546434772,
    8505806.597449332,
    7272245.97545039,
    9785009.118050914,
    8023968.06734304,
    8370252.207796229,
    7616759.566780881,
    9758307.209897608,
    9549180.910991326,
    11536569.38123997,
    17198302.677047078
]
```
</details>

REPL:

https://svelte.dev/repl/47f4bf4cb3914fad9d8cd1d8bd13ada5?version=3.56.0

I'll look into this further when I can, but I'm hoping someone might have a better idea what's going on. Thank you!

donmccurdy:
Opened a PR with a (failing) unit test —

[...]

PR #705, opened 2023-03-14
Add failing test for consistency of quantile() array vs. scalar 'p' parameter

donmccurdy:
Context: 

- #704


MohammedAlkindi:
Data point for whether this draft is still needed: the inconsistency it guards against was fixed in 7.9.2 (#805), and `main` now carries a regression test on essentially this dataset — `test/quantile.test.js`, "array of quantiles is independent of which quantiles are requested (#704)". Running this draft's exact fixture on today's `main` (8449cf5):

```js
ss.quantile(input, [0, 0.05, 0.95, 1.0]);             // [0, 3641749.5, 11043884.9, 17198302]
[0, 0.05, 0.95, 1.0].map((p) => ss.quantile(input, p)); // identical, element for element
```

Array and scalar agree. The draft's literal expectations (`[0, 3569031, 11183760, 17198302]`) were written against the pre-7.8.9 interpolation; since #783 switched `quantile` to type-7, neither call form returns those values anymore, so the test as written would fail on values while the property it exists to check now holds and is pinned by the #704 test. It is also written for `tap` (`t.test`/`t.same`/`t.end`), which #807 removed in favor of `node:test`, so the hunk no longer fits the rewritten suite. Seems safe to close.

donmccurdy:
Thanks for the update @MohammedAlkindi!

PR #717, opened 2023-07-11
Faster rounding in `cumulativeStdNormalProbability`

### LINKED ISSUE #716: why is a math lib returning strings?
cumulativeStdNormalProbability uses toFixed(), which returns a string.  This is not how you round.

mourner:
Closes #716.


tmcw (approved):
Thanks Volodymyr!

PR #718, opened 2023-09-14
Replace Math.min and Math.max with min and max

harunorimurata:
Just replaced `Math.min()` and `Math.max()` with the custom `min()`, `max()`.


tmcw:
I like the idea of using built-ins like this, but I think in the case where you have known arity, `Math.min` and `Math.max` are going to be a lot faster than creating an array and iterating.

PR #737, opened 2024-10-02
fix website link in .d.ts files

kalijonn:
Update website url in .d.ts files


kalijonn:
Not sure if you prefer an issue first. looked like a small enough change to just create a PR.

tmcw (approved):
Thank you!

Issue #789, opened 2026-06-01
Add Weighted Statistics Formula


rrbharath:
Add the formulas for the following:

1. weighted mean
2. weighted quantile
3. weighted standard deviation
4. weighted variance

MohammedAlkindi:
All four of these shipped in #790 (merged 2026-06-04), which referenced this issue but without a closing keyword, so it stayed open. Verified just now at current `main` (7063073):

```js
import * as ss from "simple-statistics";
ss.weightedMean([1, 2, 3], [1, 1, 2]);              // 2.25
ss.weightedQuantile([1, 2, 3], [1, 1, 2], 0.5);     // 2
ss.weightedStandardDeviation([1, 2, 3], [1, 1, 2]); // 0.82915619758885
ss.weightedVariance([1, 2, 3], [1, 1, 2]);          // 0.6875
```

Looks closeable as completed by #790.

PR #790, opened 2026-06-01
feat(stats): add weighted statistics formulas

rrbharath:
Add weighted statistic formulas.

Addresses issue: https://github.com/simple-statistics/simple-statistics/issues/789

### DIFF
--- a/.changeset/weighted-statistics.md
+++ b/.changeset/weighted-statistics.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": minor
+---
+
+Add weighted mean, weighted variance, weighted standard deviation, and weighted quantile.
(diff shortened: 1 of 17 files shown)


tmcw (approved):
Great - I think these methods do beg the question of whether the type annotations should be in typescript natively because the JSDoc types here are weaker than the typescript types.

tmcw:
Okay, there is an issue - the final quantile line does not have code coverage:

https://github.com/simple-statistics/simple-statistics/blob/db27d77615aa30237f7869cb8468a392dce7f64f/src/weighted_quantile.js#L63-L63

tmcw (changes requested):
Great, but needs code coverage of the final quantile line.

PR #800, opened 2026-06-24
fix: gamma() no longer infinitely recurses for 0 < n < 1

spokodev:
## Problem

`gamma(n)` throws `RangeError: Maximum call stack size exceeded` for every non-integer argument in the open interval `(0, 1)`:

```js
ss.gamma(0.5);  // RangeError (expected ~1.7724538509055159, sqrt(pi))
ss.gamma(0.25); // RangeError
ss.gamma(0.9);  // RangeError
```

These inputs are inside the documented domain. The JSDoc states gamma is "defined for all real n except zero and negative integers (where NaN is returned)". Integer arguments, arguments greater than or equal to 1, and negative non-integers all work; only `(0, 1)` overflows the stack.

## Root cause

For an argument `n` with `0 < n < 1`, after `n--` the value lands in `(-1, 0)`, so the reflection branch runs and recurses on `gamma(-n)`, i.e. `gamma(1 - originalN)`. That reflected argument is also in `(0, 1)`, so it reaches neither the integer branch nor the Nemes expansion (which needs the decremented `n` to be non-negative, meaning the original argument is at least 1). The recursion has no base case for `(0, 1)`. For `0.5` it is a tight two-cycle: `gamma(0.5)` -> `gamma(0.5)` -> ...

## Fix

Intercept arguments in `(0, 1)` before the decrement and lift them into the supported domain using the recurrence `gamma(n) = gamma(n + 1) / n`, where `n + 1` is in `(1, 2)` and resolves through the Nemes branch. Arguments outside `(0, 1)` are untouched.

```js
if (n > 0 && n < 1) {
    return gamma(n + 1) / n;
}
```

## Before / after

| call | before | after |
|---|---|---|
| `gamma(0.5)` | RangeError | 1.7759207786637015 |
| `gamma(0.25)` | RangeError | 3.7281456682241094 |
| `gamma(0.9)` | RangeError | 1.0687385151536282 |
| `gamma(5)` | 24 | 24 |
| `gamma(11.54)` | 13098426.039156161 | 13098426.039156161 |
| `gamma(-42.5)` | -3.419793520724856e-52 | -3.419793520724856e-52 |

This is strictly a termination fix. The returned values reflect the existing Nemes approximation's accuracy at small arguments and are not a new precision claim; the recurrence identity `gamma(0.5) === gamma(1.5) / 0.5` holds exactly.

## Tests

Added three cases to `test/gamma.test.js`: finiteness for `0.5`, `0.25`, `0.9`; the exact recurrence identity for `0.5`; and `gamma(0.5)` approximately `sqrt(pi)`. They fail (RangeError) on the current code and pass with the fix. Full suite is green (779/779).

### DIFF
--- a/.changeset/gamma-recursion-sub-one.md
+++ b/.changeset/gamma-recursion-sub-one.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+fix: gamma() no longer infinitely recurses for 0 < n < 1
--- a/src/gamma.js
+++ b/src/gamma.js
@@ -27,6 +27,18 @@ function gamma(n) {
         }
     }
 
+    if (n > 0 && n < 1) {
+        // Arguments in (0, 1) would otherwise recurse forever through
+        // the reflection branch below, because the reflected argument
+        // stays in (0, 1) and never reaches the Nemes expansion.
+        // Lift into the supported domain via the recurrence
+        // gamma(n) = gamma(n + 2) / (n * (n + 1)). Stepping up by two
+        // lands the evaluation on (2, 3), where the Nemes expansion is
+        // far better conditioned than just above 1, so the result is
+        // near-exact rather than merely finite.
+        return gamma(n + 2) / (n * (n + 1));
+    }
+
     // Decrement n, because approximation is defined for n - 1
     n--;
 
--- a/test/gamma.test.js
+++ b/test/gamma.test.js
@@ -24,6 +24,53 @@ test("gamma", function (t) {
         t.ok(Number.isNaN(ss.gamma(0)));
         t.end();
     });
+    t.test(
+        "gamma for argument in (0, 1) should return a finite value",
+        function (t) {
+            // Regression: these inputs previously triggered infinite
+            // recursion (RangeError: Maximum call stack size exceeded),
+            // even though gamma is documented as defined for all real n
+            // except zero and negative integers.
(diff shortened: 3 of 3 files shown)


tmcw:
Is this PR LLM-generated?

spokodev:
It is not auto-generated and dumped here. I use an AI assistant as a tool in my workflow and I am open about that. The bug is real: `gamma(n)` infinitely recurses (RangeError) for any non-integer 0 < n < 1 - e.g. `gamma(0.5)` throws - because the reflection branch recurses on `gamma(1 - n)`, which stays inside (0,1) with no terminating case.

I just pushed an update: rather than lifting the argument one step (`gamma(n+1)/n`), it lifts two (`gamma(n+2)/(n*(n+1))`). The Nemes approximation is poorly conditioned just above 1 but very good on (2,3), so the same one-line shape brings sub-1 values close to exact - gamma(0.5) is now within about 5e-6 (relative) of sqrt(pi), and gamma(0.1) drops from ~27% error to ~3e-5. Failing test first, then the fix; full suite green.

If you would rather not take LLM-assisted contributions, I understand and will close it.

tmcw:
Okay, looks good - can you add a changeset? You can do so by calling `pnpm dlx @changesets/cli`

spokodev:
Done - added a patch changeset.

tmcw (approved):
Thanks!

PR #803, opened 2026-06-27
Fix quantile() returning wrong values for some requested quantile sets

### LINKED ISSUE #704: Unexpected results for multiple quantiles

chatman-media:
Fixes #704.

`quantile(x, p)` with an array `p` could return a value for one quantile that disagreed with computing that quantile on its own. The reported example:

```js
quantile(data, [0, 0.05, 0.95, 1])[1] // wrong
quantile(data, 0.05)                  // correct
```

### Cause

`multiQuantileSelect` places each requested order statistic with `quickselect`, but the recursion passed fractional indices to the child ranges (`Math.floor(indices[l])` / `Math.ceil(indices[r])`) and reused the pivot position as a shared boundary. A later, narrower `quickselect` could then re-partition a range that still contained an already-finalized position and move it. Roughly 10% of random multi-quantile requests were affected.

### Fix

Reduce the requested quantiles to the distinct integer order statistics they actually read (`floor` and `ceil` of each index, plus the endpoints), then run a multiselect that excludes each placed pivot from its child ranges, so a finalized position is never disturbed. The scalar path is unchanged.

### Verification

- New regression tests cover the reported case (array result must equal per-element scalar results) and duplicate-heavy inputs compared against a full sort. Both fail without the fix.
- Fuzzed 1M random float inputs and 1M duplicate-heavy integer inputs against a fully-sorted reference: 0 mismatches; 200k array-vs-scalar consistency checks: 0 disagreements.
- Full suite (785 tests), `tsc`, and `biome` pass.

### DIFF
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -53,29 +53,34 @@ function quantileSelect(arr, k, left, right) {
 }
 
 function multiQuantileSelect(arr, p) {
-    const indices = [0];
+    // Collect every integer order statistic that the requested quantiles need.
+    // Linear interpolation (type=7) reads x[floor(idx)] and x[ceil(idx)], so both
+    // surrounding positions must be placed. Deduplicate to integer positions
+    // (including the endpoints) so each is selected with an unambiguous index.
+    const positions = new Set([0, arr.length - 1]);
     for (let i = 0; i < p.length; i++) {
-        indices.push(quantileIndex(arr.length, p[i]));
+        const idx = quantileIndex(arr.length, p[i]);
+        positions.add(Math.floor(idx));
+        positions.add(Math.ceil(idx));
     }
-    indices.push(arr.length - 1);
-    indices.sort(compare);
+    const indices = Array.from(positions).sort(compare);
 
-    const stack = [0, indices.length - 1];
-
+    // Recursively partition around the median requested position, then the two
+    // halves. The just-placed pivot is excluded from the child ranges so that a
+    // later, narrower quickselect cannot disturb an already-finalized position.
+    const stack = [0, indices.length - 1, 0, arr.length - 1];
     while (stack.length) {
-        const r = Math.ceil(stack.pop());
-        const l = Math.floor(stack.pop());
-        if (r - l <= 1) continue;
+        const right = stack.pop();
+        const left = stack.pop();
+        const r = stack.pop();
+        const l = stack.pop();
+        if (l > r) continue;
 
-        const m = Math.floor((l + r) / 2);
-        quantileSelect(
-            arr,
-            indices[m],
-            Math.floor(indices[l]),
-            Math.ceil(indices[r])
-        );
+        const m = (l + r) >> 1;
+        quickselect(arr, indices[m], left, right);
 
-        stack.push(l, m, m, r);
+        stack.push(l, m - 1, left, indices[m] - 1);
+        stack.push(m + 1, r, indices[m] + 1, right);
     }
 }
 
--- a/src/quantile.js
+++ b/src/quantile.js
@@ -56,8 +56,8 @@ function multiQuantileSelect(arr, p) {
     // Collect every integer order statistic that the requested quantiles need.
     // Linear interpolation (type=7) reads x[floor(idx)] and x[ceil(idx)], so both
     // surrounding positions must be placed. Deduplicate to integer positions
-    // (including the endpoints) so each is selected with an unambiguous index.
-    const positions = new Set([0, arr.length - 1]);
+    // so each is selected with an unambiguous index.
+    const positions = new Set();
     for (let i = 0; i < p.length; i++) {
         const idx = quantileIndex(arr.length, p[i]);
         positions.add(Math.floor(idx));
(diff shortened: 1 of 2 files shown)


mourner:
@chatman-media oh, really nice catch — thank you for a thorough, thoughtful PR! Can you also run some benchmarks just to make sure performance of multi quantile doesn't regress (I assume it doesn't)?

chatman-media:
Ran some benchmarks — comparing this branch against main on the same random inputs (sizes 1e3 → 1e6, warmed up, medians over many iterations), requesting both a small set `[0, .05, .25, .5, .75, .95, 1]` and a dense 101-quantile set (`0..1` step `.01`).

Good news: no regression. Multi-quantile is basically at parity and actually a touch faster on larger arrays, which makes sense since deduping to distinct integer order statistics means slightly fewer quickselect passes. A few representative medians:

```
 n=10k,  7 quantiles:   0.185ms -> 0.153ms   (~0.83x)
 n=1M,   7 quantiles:   16.2ms  -> 15.7ms    (~0.97x)
 n=1M,   101 quantiles: 46.1ms  -> 44.8ms    (~0.97x)
```

The only spot where the fix is marginally slower is tiny arrays — `n=1000` with 7 quantiles is ~1.15x (about 8.3µs → 9.5µs, so ~1µs of constant Set/dedup overhead), and that flips to a win by 10k elements. Scalar `quantile(x, 0.5)` is unchanged. Happy to share the script if useful.

src/quantile.js
```diff
@@ -53,29 +53,34 @@ function quantileSelect(arr, k, left, right) {
 }
 
 function multiQuantileSelect(arr, p) {
-    const indices = [0];
+    // Collect every integer order statistic that the requested quantiles need.
+    // Linear interpolation (type=7) reads x[floor(idx)] and x[ceil(idx)], so both
+    // surrounding positions must be placed. Deduplicate to integer positions
+    // (including the endpoints) so each is selected with an unambiguous index.
+    const positions = new Set([0, arr.length - 1]);
```
mourner: Looks like this is a leftover:
```suggestion
    // so each is selected with an unambiguous index.
    const positions = new Set();
```

src/quantile.js
```diff
@@ -53,29 +53,34 @@ function quantileSelect(arr, k, left, right) {
...
-            indices[m],
-            Math.floor(indices[l]),
-            Math.ceil(indices[r])
-        );
+        const m = (l + r) >> 1;
+        quickselect(arr, indices[m], left, right);
 
-        stack.push(l, m, m, r);
+        stack.push(l, m - 1, left, indices[m] - 1);
+        stack.push(m + 1, r, indices[m] + 1, right);
```
mourner: The stack version doesn't read too nicely — maybe worth replacing with a compact recursive version? Perf should be the same.

```js
function multiselect(arr, indices, lo, hi, left, right) {
    if (lo > hi) return;
    const mid = (lo + hi) >> 1;
    quickselect(arr, indices[mid], left, right);
    multiselect(arr, indices, lo, mid - 1, left, indices[mid] - 1);
    multiselect(arr, indices, mid + 1, hi, indices[mid] + 1, right);
}
```

mourner (changes requested):
Looks good apart from a few nits

chatman-media:
Applied both — dropped the leftover comment/seed and switched the stack loop to your recursive `multiselect`. Reads much nicer, and it stays behaviour-identical (cross-checked against a full sort). Thanks!

tmcw:
@mourner is the expert here so once it's good with him, I'll merge!

mourner (approved):
Nice! BTW @tmcw does it need some new file in the `changeset` folder? I don't know how that works and it's not in `CONTRIBUTING.md`...

tmcw:
Ah, yeah - I need to add that. Running `pnpm @changesets/cli` should do the trick

tmcw:
Merged with #805

PR #804, opened 2026-06-29
fix: permutationTest uses original sample sizes for unequal-length samples

mahirhir:
`permutationTest` builds its null distribution by shuffling the pooled data and re-splitting it, but it splits at the midpoint of the combined array:

```js
const allData = sampleX.concat(sampleY);
const midIndex = Math.floor(allData.length / 2);
```

The observed statistic is `mean(sampleX) - mean(sampleY)`, computed on the real group sizes, while every permutation statistic is `mean(left) - mean(right)` on two equal halves. Those line up only when `sampleX` and `sampleY` have the same length. For unequal samples the permutation distribution is built from the wrong group sizes, so the resulting p-value is not meaningful.

Concrete case: `permutationTest([10, 10, 10], [0], "two_side")`. The observed difference is `10`. The pool `[10, 10, 10, 0]` gets split 2-vs-2, whose mean difference is at most `5`, so no permutation reaches `10` and the p-value collapses to `0`. The correct value is `0.25`, since only the arrangement that isolates the `0` is as extreme as the observed difference.

The fix splits each shuffled pool at `sampleX.length`, so the two groups keep their original sizes. When the samples are already equal length this is the same as the old midpoint (`floor(2n / 2) === n`), so existing results do not change. I added a regression test for the unequal-length case.

`tap test/*.test.js` passes (782 assertions) and `biome check` is clean. The midpoint split has been there since the function was added in #288.

### DIFF
--- a/.changeset/permutation-test-unequal-sizes.md
+++ b/.changeset/permutation-test-unequal-sizes.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+fix: permutationTest now keeps the original sample sizes when permuting unequal-length samples
--- a/src/permutation_test.js
+++ b/src/permutation_test.js
@@ -54,13 +54,17 @@ function permutationTest(sampleX, sampleY, alternative, k, randomSource) {
 
     // combine datsets so we can easily shuffle later
     const allData = sampleX.concat(sampleY);
-    const midIndex = Math.floor(allData.length / 2);
+    // Each permutation reassigns the pooled values into two groups that keep
+    // the original sample sizes. Splitting at the midpoint only matches when
+    // both samples are the same length; for unequal samples it built the
+    // permutation distribution from the wrong group sizes.
+    const splitIndex = sampleX.length;
 
     for (let i = 0; i < k; i++) {
         // 1. shuffle data assignments
         shuffleInPlace(allData, randomSource);
-        const permLeft = allData.slice(0, midIndex);
-        const permRight = allData.slice(midIndex, allData.length);
+        const permLeft = allData.slice(0, splitIndex);
+        const permRight = allData.slice(splitIndex, allData.length);
 
         // 2.re-calculate test statistic
         const permTestStatistic = mean(permLeft) - mean(permRight);
--- a/test/permutation_test.test.js
+++ b/test/permutation_test.test.js
@@ -67,5 +67,20 @@ test("permutation test", function (t) {
         }
     );
 
+    t.test("handles samples of unequal length", function (t) {
+        // Each permutation keeps the original group sizes (3 and 1). The
+        // observed difference is mean([10, 10, 10]) - mean([0]) = 10, which a
+        // permutation can match only when the single 0 lands in the second
+        // group, one quarter of the time. Splitting the pool in half instead
+        // compared this against 2-vs-2 permutations, whose mean difference can
+        // never reach 10, so the p-value collapsed to 0.
+        const seeded = new Random.Random(Random.MersenneTwister19937.seed(42));
+        const p = ss.permutationTest([10, 10, 10], [0], "two_side", 10000, () =>
+            seeded.real(0, 1)
+        );
+        t.ok(Math.abs(p - 0.25) < 0.05);
+        t.end();
+    });
+
     t.end();
 });
--- a/test/permutation_test.test.js
+++ b/test/permutation_test.test.js
@@ -51,4 +51,18 @@ describe("permutation test", function () {
             ss.permutationTest([1, 69, 420], [42, 42, 42], "one-tailed");
         });
     });
+
+    it("handles samples of unequal length", function () {
+        // Each permutation keeps the original group sizes (3 and 1). The
+        // observed difference is mean([10, 10, 10]) - mean([0]) = 10, which a
+        // permutation can match only when the single 0 lands in the second
+        // group, one quarter of the time. Splitting the pool in half instead
+        // compared this against 2-vs-2 permutations, whose mean difference can
+        // never reach 10, so the p-value collapsed to 0.
+        const seeded = new Random.Random(Random.MersenneTwister19937.seed(42));
+        const p = ss.permutationTest([10, 10, 10], [0], "two_side", 10000, () =>
+            seeded.real(0, 1)
+        );
+        assert.ok(Math.abs(p - 0.25) < 0.05);
+    });
 });


MohammedAlkindi (commented):
The diagnosis and the fix in `src/permutation_test.js` are right, and I think this is blocked on something small and mechanical rather than on anything about the statistics.

**The fix is correct.** Splitting at `sampleX.length` is the right invariant: the observed statistic is computed on the real group sizes, so the permutation statistics have to be too, otherwise the null distribution answers a different question than the one being asked. I worked the example by hand and agree with 0.25. Pool `[10, 10, 10, 0]`, groups of 3 and 1, four arrangements; the `0` lands alone in the second group in one of them, giving `10 - 0 = 10`, which matches the observed difference. In the other three the `0` sits in the first group, giving `20/3 - 10 = -3.33`, which is not as extreme. One of four is 0.25. And `floor(2n / 2) === n`, so equal-length samples are genuinely unaffected — that is what makes this safe to land.

**What is blocking it: the test is written for a test runner the repo no longer has.** #807 ("Modernize test setup, switch the package to ESM") landed on 2026-07-03, four days after this PR was opened, and moved every test file to `node:test`. `test/permutation_test.test.js` on `main` now starts:

```js
import assert from "node:assert/strict";
import { describe, it } from "node:test";
import * as Random from "random-js";

const random = new Random.Random(Random.MersenneTwister19937.seed(0));
...
function rng() {
    return random.real(0, 1);
}

describe("permutation test", function () {
    it("P-value of identical distributions being different should be 1", function () {
```

and `package.json` is now `node --test --experimental-test-isolation=none test/*.test.js`, with `tap` gone from `devDependencies`. So `t.test(...)`, `t.ok(...)` and `t.end()` no longer resolve, and the hunk will not apply against the rewritten file. The PR body's `tap test/*.test.js passes (782 assertions)` was accurate when written and is just stale now.

Ported, the added case would be roughly:

```js
    it("handles samples of unequal length", function () {
        // Each permutation keeps the original group sizes (3 and 1). The
        // observed difference is mean([10, 10, 10]) - mean([0]) = 10, which a
        // permutation can match only when the single 0 lands in the second
        // group, one quarter of the time. Splitting the pool in half instead
        // compared this against 2-vs-2 permutations, whose mean difference can
        // never reach 10, so the p-value collapsed to 0.
        const seeded = new Random.Random(Random.MersenneTwister19937.seed(42));
        const p = ss.permutationTest([10, 10, 10], [0], "two_side", 10000, () =>
            seeded.real(0, 1)
        );
        assert.ok(Math.abs(p - 0.25) < 0.05);
    });
```

`random-js` is still a devDependency and the file still imports it, so the seeded-source pattern carries over unchanged. Worth deciding whether to use the file's existing module-level `rng` helper or keep the locally seeded generator — a local one is better here, since it makes this case independent of how many draws earlier tests consumed.

Two smaller notes, neither blocking:

- The `0.05` tolerance is doing nothing, since the source is seeded and the result is deterministic. That is fine and arguably more robust to a `random-js` version change, but it is worth a word in the comment so a later reader does not think the test is flaky by design.
- `.changeset/permutation-test-unequal-sizes.md` is present and correctly shaped, which is more than most PRs here manage.

One thing I would leave alone: `src/permutation_test.js` lines 77, 83 and 90 iterate `i <= k` over a length-`k` array. It reads like an off-by-one, but `testStatDsn[k]` is `undefined` and every comparison against `undefined` is `false`, so the counts are unaffected. Not worth widening this PR for.

Context for the author: I have no commit rights here, this is just a passing read. I found the same midpoint bug independently while looking at this file and stopped when I saw you had already filed it — the fix is yours and it is the right one. It would be a shame for it to lapse over a test-runner rename.

MohammedAlkindi:
The conflict is the test harness, not your change. The case you added is written against `tap` (`t.test` / `t.ok` / `t.end`), and #807 replaced tap with `node:test` on 2026-07-03, rewriting `test/permutation_test.test.js` underneath you.

Porting it is mechanical: the file now opens with `import assert from "node:assert/strict"` and `import { describe, it } from "node:test"`, `t.ok(x)` becomes `assert.ok(x)`, and the seeded `random-js` pattern you want is already at the top of that file.

For what it's worth, the underlying bug reproduces here: `permutationTest([9, 10], [1..10], "two_side", 200000)` returns 0.0174, against an exact 0.1364 enumerated over all C(12,2) splits at the true 2/10 group sizes.

MohammedAlkindi:
Checked the fix itself this time, not just the conflict.

Applying `splitIndex = sampleX.length` to current `main` and comparing against exact enumeration over all C(n,k) splits: 2-vs-10 moves from 0.0173 to 0.1362 against an exact 0.1364, and 3-vs-9 from 0.0217 to 0.0558 against 0.0546. Equal-length input stays where it was, 4-vs-4 giving 0.0284 before and 0.0288 after against an exact 0.0286.

So the change is correct and it leaves the equal-length path alone. The only thing standing between this and a clean run is the harness port I mentioned above.

mahirhir:
Rebased onto current main and ported the added case to node:test as you described: tap's t.ok became assert.ok and the seeded random-js pattern now matches the top of the file. Full suite locally is 314 pass, 0 fail, and biome is clean on both changed files. Thanks for verifying the fix against exact enumeration on 2-vs-10, 3-vs-9, and 4-vs-4; nothing in the src change moved since your check, the only delta is the harness port.

MohammedAlkindi:
Port looks right, and the numbers hold. I re-ran it on your branch at `3806029`, rebased against `a8c348c`: 314 passing, 0 failing, which matches what you got.

The `src/permutation_test.js` diff is unchanged from what I checked on 24 August, still just `midIndex` to `splitIndex = sampleX.length`, so the exact-enumeration results from that comment still apply and I did not need to redo them.

The new test's reasoning is right too. Pooled `[10, 10, 10, 0]` split 3-and-1, the observed difference is 10, and a permutation reaches it only when the single `0` lands in the second group, which is one placement in four. So 0.25 is the exact value, not an approximation, and the 0.05 band is generous.

One thing worth flagging for whoever picks this up: the check suite on this PR reports `action_required` with **0** runs, so no job has executed. That is the first-time-contributor approval gate, not a failing build, and it is why GitHub shows the PR as unstable. It needs a maintainer to click "Approve and run workflows", and pushing again re-arms it rather than clearing it.

tmcw (approved):
Thanks!

PR #807, opened 2026-07-03
Modernize test setup, switch the package to ESM

mourner:
Noticed that the project still used the super ancient `tap` dependency for tests that is long obsolete — it gets some updates but it's a bloated monster that is fully superceded by native Node test runner. A good illustration of this is that removing the dependency also removes 2.2K lines from the pnpm lockfile. `npm test` takes 3 seconds (actual unit tests are ~120ms).

Also removes some ancient leftovers no longer used, switched the package to `"type": "module"` which actually reflects how it's structured in the source (while preserving compatibility via `exports` field), and fixed an issue where the version in the readme was no longer being updated on release.


tmcw (approved):
Awesome! I have definitely appreciated that `tap` has been maintained for many years but yeah, ~94MB with dependencies and the `node:test` system working pretty well means there's really no reason for it.

Thanks! Hope you're doing well!

PR #809, opened 2026-07-08
fix(wilcoxon): reset tied ranks between separate tie groups

spokodev:
The tie averaging in `wilcoxonRankSum` reused the `tiedRanks` buffer when a value change closed a tie group. The buffer was averaged but never reset, so a later tie appended its ranks to the stale group and the average spanned unrelated positions. This shows up whenever the pooled sample has more than one tied group, for example `wilcoxonRankSum([1, 1], [2, 2])` returned 5 instead of 3.

The fix resets `tiedRanks` to the current element after averaging a closed group, matching the reset already done in the single-element branch. Adds a regression test for separate tied groups.


tmcw:
Thank you! Merged with #816

PR #810, opened 2026-07-08
fix(bayes): accumulate scores across all item properties

spokodev:
`BayesianClassifier.score` reset the per-category odds object on every property of the scored item, so all but the last property were discarded and multi-property items were scored on a single attribute. After training two `animal` items sharing `species: "Cat"` with different colors, scoring `{ species: "Cat", color: "white" }` returned `{ animal: 0.5 }` instead of `{ animal: 1.5 }`.

The fix initializes each category's odds object once per `score` call so every property contributes to the tallied sum. Existing single-property tests are unaffected. Adds a multi-property regression test.

### DIFF
--- a/src/bayesian_classifier.js
+++ b/src/bayesian_classifier.js
@@ -83,8 +83,11 @@ class BayesianClassifier {
             const v = item[k];
             for (category in this.data) {
                 // Create an empty object for storing key - value combinations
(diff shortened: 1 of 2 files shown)


tmcw (approved):
Thanks! LGTM

PR #811, opened 2026-07-26
fix: discrete distributions no longer truncate on intermediate overflow

binggao1230:
`poissonDistribution` and `binomialDistribution` return a table that stops short of its own mean and ends on a non-finite cell, for arguments well inside their documented domain:

```js
const cells = ss.poissonDistribution(200);
cells.length; // 135, and the distribution peaks at 200
cells[134]; // Infinity
ss.sum(cells); // NaN - the finite cells are 0.000045% of the distribution

ss.binomialDistribution(10000, 0.5); // 135 cells, all zero except a final NaN
```

The first failing arguments are `lambda >= 111` and, at `p = 0.5`, `trials >= 1021`. Nothing is thrown and `undefined` (the documented invalid-input signal) is not returned, because the input is valid.

## Root cause

Both functions build each cell from factors that leave floating-point range long before the cell itself does: `Math.pow(lambda, x) / factorialX` for Poisson, `binomialCoefficient * Math.pow(probability, x) * Math.pow(1 - probability, trials - x)` for binomial. `Math.pow(200, 134)` is `Infinity` while the probability it belongs to is 1.5e-7. Past `lambda = 745` the leading `Math.exp(-lambda)` underflows to zero as well, so the cell is `0 * Infinity`, i.e. `NaN`.

Both loops then share this stopping rule:

```js
} while (cumulativeProbability < 1 - epsilon);
```

`Infinity < 0.9999` and `NaN < 0.9999` are both false, so the overflow *is* the exit condition. The loop reports that it covered the distribution and hands back the truncated table.

## Extent

I swept both functions against `scipy.stats` (checked in turn against mpmath at 50 digits, and against the Hines & Montgomery values the tests already pin) over 28 values of `lambda` and 124 `(trials, probability)` pairs. Counting a table as wrong if it holds a non-finite cell or covers less than `1 - epsilon` of the distribution: **14/28 Poisson and 44/124 binomial before, 0 and 0 after.**

| call | cells returned | share of the distribution | mode |
| --- | --- | --- | --- |
| `poissonDistribution(150)` | 143 | 27% | 150 |
| `poissonDistribution(200)` | 135 | 0.000045% | 200 |
| `poissonDistribution(1000)` | 104 | 5.7e-288% | 1000 |
| `binomialDistribution(1021, 0.5)` | 497 | 19% | 511 |
| `binomialDistribution(2000, 0.5)` | 229 | 4.2e-294% | 1000 |
| `binomialDistribution(10000, 0.5)` | 135 | 0% | 5000 |
| `binomialDistribution(1000000, 0.0001)` | 68 | 0.029% | 100 |

`bernoulliDistribution` is a closed form and is unaffected. `chiSquaredGoodnessOfFit` calls `poissonDistribution` and inherits the truncated table for any sample with a mean of 111 or more.

## Fix

Each cell is taken through its logarithm using `gammaln`, which the library already ships and exports for exactly this problem ("useful for values of n too large for the normal gamma function"):

```js
// poisson
Math.exp(x * Math.log(lambda) - lambda - gammaln(x + 1));

// binomial
Math.exp(
    gammaln(trials + 1) - gammaln(x + 1) - gammaln(trials - x + 1) +
        x * Math.log(probability) + (trials - x) * Math.log(1 - probability)
);
```

The cumulative probability is now checked after the loop rather than trusted: a table that did not reach `1 - epsilon` is not a distribution, so `undefined` is returned instead of part of one. That also covers `poissonDistribution(NaN)` and `binomialDistribution(5, NaN)`, which used to return `[NaN]` — `NaN` slips through the existing input checks.

The loops need a bound now that they run all the way to the mode. `binomialDistribution` stops after `trials`, since more successes than trials are impossible, and both stop at a million cells: `poissonDistribution(1e9)` would otherwise try to build a billion-cell array, and instead returns `undefined` in about 20ms. `poissonDistribution(900000)` still returns its full 903531-cell table.

## Accuracy

Evaluating cells independently through logarithms gives up some precision where the product form worked. Worst relative error against the oracle, before -> after: `lambda <= 110`, 1.8e-13 -> 2.3e-13; `binomialDistribution(1000, 0.5)`, 1.1e-13 -> 1.6e-12; `binomialDistribution(10000, 0.001)`, 1.9e-14 -> 1.7e-11. Everything past those points used to be a truncated table and is now good to 4e-9 or better at a million trials. The worst of it is seven orders of magnitude inside the `epsilon` these functions already use as their stopping rule.

## Tests

Nine cases across the two test files, all failing on `main` except the `probability` 0 and 1 pin, which is there to hold the existing degenerate output. They cover: every cell finite and the table summing to at least `1 - epsilon`, for nine values of `lambda` and six `(trials, probability)` pairs; the peak landing at the mean for `poissonDistribution(200)` and `binomialDistribution(10000, 0.5)`; thirteen reference cells computed at 50 digits with mpmath and cross-checked against `scipy.stats`; and the `undefined` cases. Full suite 287 passing, `biome` and `documentation lint` clean.

### DIFF
--- a/.changeset/discrete-pmf-overflow.md
+++ b/.changeset/discrete-pmf-overflow.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Fix `poissonDistribution()` and `binomialDistribution()` returning a truncated table ending in `Infinity` or `NaN`. Both built each cell from a power and a factorial (or a binomial coefficient) that leave floating-point range long before their product does, so the cumulative-probability stopping rule was satisfied by a non-finite sum instead of by the distribution: `poissonDistribution(200)` returned 135 cells covering 0.000045% of the distribution, and `binomialDistribution(10000, 0.5)` returned 135 cells covering none of it. Cells are now taken through `gammaln`, and a table that cannot be completed returns `undefined` rather than a partial one.
--- a/src/binomial_distribution.js
+++ b/src/binomial_distribution.js
@@ -1,4 +1,46 @@
 import epsilon from "./epsilon.js";
+import gammaln from "./gammaln.js";
+
+// A distribution over more trials than this would need a table we cannot
+// allocate. Past that we report a distribution we cannot produce rather than
+// a partial one.
+const maxCells = 1e6;
+
+/**
+ * The probability of exactly `x` successes, taken through its logarithm: the
+ * binomial coefficient and the two powers each leave floating-point range long
+ * before the product they form does, so computing them separately loses cells
+ * the distribution needs.
+ *
+ * @private
+ * @param {number} trials number of trials
+ * @param {number} probability probability of success in one trial
+ * @param {number} x number of successes
+ * @returns {number} probability of exactly `x` successes
+ */
+function probabilityOfSuccesses(trials, probability, x) {
+    // The logarithms below are undefined at the ends of the probability range,
+    // where the outcome is certain anyway.
+    if (probability === 0) {
+        if (x === 0) {
+            return 1;
+        }
+        return 0;
+    }
+    if (probability === 1) {
+        if (x === trials) {
+            return 1;
+        }
+        return 0;
+    }
+    return Math.exp(
+        gammaln(trials + 1) -
+            gammaln(x + 1) -
+            gammaln(trials - x + 1) +
+            x * Math.log(probability) +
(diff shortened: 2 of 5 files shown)


tmcw:
Merged with https://github.com/simple-statistics/simple-statistics/pull/820 - thank you!

PR #812, opened 2026-07-26
fix: chiSquaredGoodnessOfFit no longer reports a fit it never measured

binggao1230:
`chiSquaredGoodnessOfFit` reports a fit whenever its arithmetic produces `NaN`, because `criticalValue < NaN` is `false` — the same answer a good fit gives. Two things produce that `NaN`: collapsing a sparse class pops the last class instead of the one just merged, which discards observations and double counts others (six of eleven disappear; five survive on this suite's own "gaps" fixture), and an observation past the end of the hypothesized distribution has no expected frequency to pair with. A significance level outside the table's eleven columns reads as `undefined` and reports a fit too, and degrees of freedom outside the table threw a `TypeError`.

```js
// 25 zeros, 10 ones, 10 twos, 10 threes, 5 fours
ss.chiSquaredGoodnessOfFit(data, ss.poissonDistribution, 0.05);              // true, rejects
ss.chiSquaredGoodnessOfFit(data.concat([30]), ss.poissonDistribution, 0.05); // false, accepts
```

One extreme outlier is stronger evidence against Poisson, not weaker (exact p goes from 1.5e-4 to 3.0e-6), and it flipped the verdict to "fits".

Observations past the distribution now fold into its last class, the collapse removes the class it merged, and a lookup the table cannot answer raises a descriptive error. The three existing fixtures keep their values; I checked the corrected statistics against exact chi-squared p-values computed with mpmath. 285 tests passing.

### DIFF
--- a/.changeset/chi-squared-unchecked-lookups.md
+++ b/.changeset/chi-squared-unchecked-lookups.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Fix `chiSquaredGoodnessOfFit()` reporting a fit it never measured. Collapsing a sparse class dropped the last class instead of the one that had just been merged, so observations were discarded and others double counted; an observation past the end of the hypothesized distribution had no expected frequency at all, which made the statistic `NaN`; and `criticalValue < NaN` is `false`, the same answer the function gives for a good fit. Degrees of freedom and significance levels the table does not cover now raise a descriptive error instead of a `TypeError` or a silent `false`.
--- a/src/chi_squared_goodness_of_fit.js
+++ b/src/chi_squared_goodness_of_fit.js
@@ -68,16 +68,37 @@ function chiSquaredGoodnessOfFit(data, distributionType, significance) {
         }
     }
 
+    if (expectedFrequencies.length === 0) {
+        throw new Error(
+            "chiSquaredGoodnessOfFit requires distributionType to return a distribution for the mean of the data"
+        );
+    }
+
+    // The hypothesized distribution can stop short of the largest observation,
+    // leaving those observations with no expected frequency to be compared
+    // against. Fold them into the last class the distribution does cover, the
+    // same treatment the collapse below gives to sparse classes, so that the
+    // two histograms describe the same classes.
+    const lastClass = expectedFrequencies.length - 1;
+    for (let k = observedFrequencies.length - 1; k > lastClass; k--) {
+        observedFrequencies[lastClass] += observedFrequencies[k];
+        observedFrequencies.pop();
+    }
+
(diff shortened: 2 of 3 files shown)


MohammedAlkindi (commented):
I checked this out as the test-merge against today's `main` (8449cf5) and re-verified each claim in the body independently. Everything holds.

Ran the full gate on the merge ref — `tsc --noEmit`, `biome check`, `documentation lint`, `node --test` — on Windows 11 / Node 24 / pnpm 11.17: **301/301 pass, exit 0.** On `main` the three failure modes reproduce exactly as described:

| probe | `main` (8449cf5) | this PR (merge ref) |
|---|---|---|
| 60-pt Poisson fixture, α = 0.05 | `true` (rejects) | `true` — unchanged |
| same data + one outlier `30` | **`false` (accepts)** | `true` — the outlier now strengthens rejection |
| α = 0.03 (no table column) | **silent `false`** | throws `…no critical value tabulated for a significance of 0.03` |

The data-loss claim I checked by instrumenting the collapse loop rather than trusting the diff. On `main`, the suite's own gaps fixture `[0, 2, 3, 7, 7, 7, 7, 7, 7, 9, 10]` leaves post-collapse observed frequencies `[1, 2, 2]` — the statistic is computed on **5 of the 11 observations**. On this PR the post-collapse histogram is `[1, 2, 0, 8]`: all 11 retained, and the fixture's verdict stays `true`, now computed on the whole sample. One prose nit: the body says "five of eleven disappear" — by post-collapse sum it is five that *survive* and six that go. The defect and its direction are exactly as described either way.

I also traced why the old `pop()` only sometimes misbehaves, since the existing fixtures never caught it: popping is correct as long as every class from the tail inward collapses, so `k` stays the last index. The first time a class ≥ 3 interrupts the run, the next merge pops that surviving class instead of the one just merged — which is precisely the pattern the gaps fixture produces. That is why this survived so long with a green suite.

One thing I considered and think is fine as-is: folding out-of-support observations into the last covered class compares them against a point-mass expected frequency rather than a textbook open-ended `≥` tail class. In practice the sub-3 collapse immediately merges those tail classes into a composite whose expected value approximates the tail mass, and the new "counts observations that fall outside the hypothesized distribution" test pins the folded case directly. Not worth complicating the fix over.

Smaller things that are already right: the `k >= 1` bound removes the old `expectedFrequencies[-1]` write at `k = 0`, the error strings follow the house format (function-name-first, no trailing period), and the changeset is present. The seven added tests include the "keeps every observation when collapsing classes" case, which is the regression test this function always needed.

binggao1230:
Thanks for the careful check — corrected the count in the description.

tmcw (approved):
LGTM!

PR #814, opened 2026-08-07
fix(rank-correlation): give tied values midranks

### LINKED ISSUE #813: sampleRankCorrelation depends on row order when values are tied
`sampleRankCorrelation` returns a different value for the same paired data depending on the order the pairs are listed in, whenever either input contains tied values. Reordering rows can change the magnitude and flip the sign.

**Version:** simple-statistics 7.9.3 (npm), Node v24.18.0, Windows 11.

### Reproduction

```js
import * as ss from "simple-statistics";

// The same three paired observations {(1,1), (1,2), (2,1)},
// with the first two rows swapped.
ss.sampleRankCorrelation([1, 1, 2], [1, 2, 1]); // =>  0.5
ss.sampleRankCorrelation([1, 1, 2], [2, 1, 1]); // => -0.5
```

Three pairs is the minimum case. It gets worse with more data — over all 720 orderings of one 6-pair dataset it returns **16 distinct values**, from −0.257 to +0.657:

```js
const x = [1, 1, 2, 2, 3, 3];
const y = [1, 2, 2, 3, 1, 3];
// every permutation of the same 6 pairs:
//   simple-statistics -> 16 distinct results, min -0.257, max +0.657
//   Spearman with midranks -> one value, 0.25, for all 720
```

Two more cases that fall out of the same cause:

```js
// No monotonic association at all, by symmetry — every increase in one
// variable is matched by a decrease in the other.
ss.sampleRankCorrelation([1, 1, 2, 2], [1, 2, 1, 2]); // => 0.7999999999999999, expected 0

// y is constant, so its ranks have no variance and rho is undefined.
ss.sampleRankCorrelation([1, 2, 3, 4], [5, 5, 5, 5]); // => 1.0000000000000002, expected NaN
```

The last one seems the most likely to cause real harm: a constant column reports near-perfect monotonic association rather than signalling that the statistic does not apply.

### Cause

`src/sample_rank_correlation.js` lines 26–29 assign each value its position in the sorted array as its rank:

```js
for (let i = 0; i < xIndexes.length; i++) {
    xRanks[xIndexes[i]] = i;
    yRanks[yIndexes[i]] = i;
}
```

Tied values therefore get distinct consecutive ranks, and because `Array.prototype.sort` is stable the tiebreak is the value's original index. That makes the rank of a tied value a function of where it sits in the input array. Spearman's rho is defined on midranks, where tied values share the average of the ranks they span; with midranks the statistic depends only on the pairs.

### What I ruled out

- **Not floating point.** The differences are large and the sign changes; the control case with no ties anywhere behaves correctly and is order-invariant.
- **Not `sampleCorrelation`.** It receives the already-wrong rank vectors. Feeding it midranks computed independently gives the order-invariant answer.
- **Not previously caught by the suite**, which is why it has survived. There is no `test/sample_rank_correlation.test.js`; the only coverage is two cases inside `test/sample_correlation.test.js` (a monotonic `y = ±x²` case and the R-validated case), and both use strictly distinct values, so the tie path has no coverage at all.

Ties are the normal case for ordinal, Likert, binned, rounded, or low-cardinality data — which is much of what rank correlation is reached for in the first place.

I have a fix ready that ranks both inputs with midranks and adds the missing test file; opening it as a PR now. Results for tie-free inputs are unchanged and the existing R-validated case still passes.

MohammedAlkindi:
Closes #813.

## Summary

`sampleRankCorrelation` returned a different value for the same paired data depending on the order the pairs were listed in, whenever either input contained ties. Swapping two rows of a three-pair dataset flipped the sign; across the 720 orderings of one six-pair dataset it produced 16 distinct values. Ranking both inputs with midranks makes the statistic a property of the pairs, as Spearman's rho is defined to be.

## The cause

Lines 26–29 assigned each value its position in the sorted array as its rank:

```js
for (let i = 0; i < xIndexes.length; i++) {
    xRanks[xIndexes[i]] = i;
    yRanks[yIndexes[i]] = i;
}
```

Tied values got distinct consecutive ranks, and since `Array.prototype.sort` is stable the tiebreak was the value's original index — so the rank of a tied value was a function of where it sat in the input array. Spearman's rho is defined on midranks, where tied values share the average of the ranks they span.

## Changes

- `src/sample_rank_correlation.js` — a private `midranks` helper walks the sorted pairs, extends a run across each group of equal values, and assigns that group the average of the ranks it spans. `sampleRankCorrelation` now ranks each input through it. No change to the public API surface.
- `test/sample_rank_correlation.test.js` — new; the function had no dedicated test file.
- `.changeset/rank-correlation-midranks.md` — patch.

Two further symptoms fall out of the same fix rather than needing separate handling:

- An input with no rank variance (every value tied) now returns `NaN`, where it previously reported `1.0000000000000002`. Ranking to a constant vector gives zero standard deviation, and `sampleCorrelation` divides by it. Reporting near-perfect monotonic association for a constant column seemed the most harmful of the symptoms.
- A length mismatch now reaches `sampleCovariance`'s own "requires samples with equal lengths" error. Previously both rank arrays were sized `Array(xIndexes.length)`, so a shorter `y` left holes and the function returned `NaN` silently.

## Deliberately unchanged

Ranks stay 0-based, as before — correlation is invariant under the affine shift to 1-based ranks, and keeping the existing convention makes the diff smaller. I also left `sampleCorrelation` and `rSquared` alone; both have their own unguarded-division edge cases, but those are separate contracts and did not belong in this change.

## Testing

`pnpm test` (typecheck, biome, `documentation lint`, and the full suite) on Node v24.18.0, Windows 11:

```
ℹ tests 284
ℹ suites 74
ℹ pass 284
ℹ fail 0
```

278 before, 284 after — the six added here.

The new tests fail against the current `main` and pass with this change. Verified by stashing only `src/sample_rank_correlation.js` and running the new file against the unmodified source:

```
✖ is unchanged by reordering a larger tied dataset
  AssertionError: reordering changed the result: -0.2571428571428572 vs 0.6571428571428571
```

The existing R-validated case in `test/sample_correlation.test.js` (`cor(x, y, method = "spearman")` = 0.6484848) still passes, which is the guard that tie-free behaviour is unchanged — that dataset has no ties, so its midranks are identical to the old ranks.

I checked the results against a midrank Spearman I implemented separately rather than against remembered reference values; the expected numbers in the tests (−0.5 for the three-pair case, exactly 0 for `[1,1,2,2]` vs `[1,2,1,2]`) are derivable by hand from the midrank vectors, and I have shown that derivation in the test comments so they can be checked without running anything.

One note on running the suite from Windows: `pnpm test` fails at the biome step on a default Git for Windows clone, because `core.autocrlf` is `true`, there is no `.gitattributes`, and biome flags all 270 files for `␍`. CI is `ubuntu-latest` only so it never sees this. Cloning with `git -c core.autocrlf=false clone` avoids it. Unrelated to this change, but I would happily send a one-line `.gitattributes` separately if that is wanted.

### DIFF
--- a/.changeset/rank-correlation-midranks.md
+++ b/.changeset/rank-correlation-midranks.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Fix `sampleRankCorrelation` returning different results for the same data depending on the order the pairs are listed in. Tied values were given distinct consecutive ranks broken by original array position, rather than sharing the average of the ranks they span as Spearman's rho requires. Reordering rows could change the magnitude and even the sign of the result. Tied values now receive midranks, so the correlation depends only on the paired data. As a consequence an input with no rank variance — every value tied — now correctly returns `NaN` instead of reporting near-perfect correlation. Results for inputs with no ties are unchanged.
--- a/src/sample_rank_correlation.js
+++ b/src/sample_rank_correlation.js
@@ -1,34 +1,57 @@
 import sampleCorrelation from "./sample_correlation.js";
 
 /**
+ * Rank values, giving tied values the average of the ranks they span.
+ *
+ * Spearman's rho is defined on these midranks. Assigning tied values
+ * distinct consecutive ranks instead would make the rank of a value depend
+ * on where it happens to sit in the input array, and therefore make the
+ * correlation depend on the order the pairs are listed in.
+ *
+ * @private
+ * @param {Array<number>} values input
+ * @returns {Array<number>} ranks, in the order of the input
+ */
+function midranks(values) {
+    const order = values
+        .map((value, index) => [value, index])
+        .sort((a, b) => a[0] - b[0]);
+
+    const ranks = Array(values.length);
+    let i = 0;
+    while (i < order.length) {
+        // Extend j across every value tied with the one at i.
+        let j = i;
+        while (j + 1 < order.length && order[j + 1][0] === order[i][0]) {
+            j++;
+        }
+        const rank = (i + j) / 2;
+        for (let k = i; k <= j; k++) {
+            ranks[order[k][1]] = rank;
+        }
+        i = j + 1;
+    }
+    return ranks;
+}
+
+/**
  * The [rank correlation](https://en.wikipedia.org/wiki/Rank_correlation) is
  * a measure of the strength of monotonic relationship between two arrays
+ *
+ * Tied values share the average of the ranks they span, so the result
+ * depends only on the paired data and not on the order the pairs are given
+ * in. When either input has no rank variance the correlation is undefined
+ * and the result is `NaN`.
  *
  * @param {Array<number>} x first input
  * @param {Array<number>} y second input
  * @returns {number} sample rank correlation
+ * @example
+ * sampleRankCorrelation([1, 2, 3, 4, 5], [1, 4, 9, 16, 25]);
+ * // => 1
  */
 function sampleRankCorrelation(x, y) {
-    const xIndexes = x
-        .map((value, index) => [value, index])
-        .sort((a, b) => a[0] - b[0])
-        .map((pair) => pair[1]);
-    const yIndexes = y
-        .map((value, index) => [value, index])
-        .sort((a, b) => a[0] - b[0])
-        .map((pair) => pair[1]);
-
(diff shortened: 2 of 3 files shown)


tmcw (approved):
Cool, works for me.

Issue #469, opened 2020-06-29
Is the two-tailed pvalue correct in permutationTest?


jodorning:
When I compare the output to the same test in Python permutation_test library, it seems that your PermutationTest outputs the most frequent mean rather than the two-tailed pvalue. The one-tailed p-values are similar to those from Python permutation_test.

*SQL (BIGQUERY)*
```
CREATE TEMPORARY FUNCTION permutation_test(sampleX ARRAY<FLOAT64>, sampleY ARRAY<FLOAT64>)
 RETURNS FLOAT64
 LANGUAGE js AS
"""
  return permutationTest(sampleX, sampleY, "two_side", 5000); //simple-statistics.alternative.permutationTest(sampleX: Array<number>, sampleY: Array<number>, alternative: string, k: number): number of permutations)
"""
OPTIONS (
 library="xxxx"
);

 SELECT permutation_test([2., 5, 3, 6, 7, 2, 5], [20., 5, 13, 12, 7, 2, 2])  
```
Two-sided P = 0.133

*PYTHON*
```
import permutation_test as p

var_control = [2, 5, 3, 6, 7, 2, 5]
var_treatment = [20, 5, 13, 12, 7, 2, 2]

p_value = p.permutationtest(
    data = var_control, 
    ref_data=var_treatment,
    n_combinations_max=5000)
```
Two-sided P = 0.0999

MohammedAlkindi:
The two-sided value is correct, and the disagreement is fully attributable to the Python package. This dataset is small enough to settle exactly: with 7+7 observations there are C(14,7) = 3432 equal splits, so the whole permutation distribution can be enumerated instead of sampled.

Exact values for `[2,5,3,6,7,2,5]` vs `[20,5,13,12,7,2,2]` (observed mean difference = −31/7 ≈ −4.4286):

| quantity | exact value |
|---|---|
| two-sided, counting \|T\| ≥ \|obs\| | 460/3432 = **0.13403** |
| two-sided, doubled tail: 2·min(P(T ≤ obs), P(T ≥ obs)) | 2·230/3432 = **0.13403** |
| one-sided (less) | 230/3432 = 0.06702 |

The two standard two-sided definitions coincide here because with equal group sizes the permutation distribution is symmetric. Running this library's implementation seeded at large k converges on that exact value:

```
permutationTest([2,5,3,6,7,2,5], [20,5,13,12,7,2,2], "two_side", 200000, seededRng)
[...]

PR #828, opened 2026-08-15
docs(permutation-test): correct alternative values to 'two_side'

MohammedAlkindi:
Related: https://github.com/simple-statistics/simple-statistics/issues/469

## Summary

Calling `permutationTest` with either documented `alternative` value throws. The JSDoc names the two-sided option `two_tail` in the hypothesis list and `'two_sided' (default)` in the parameter line, but the implementation accepts only `'two_side'`:

```js
permutationTest([1, 2, 3], [4, 5, 6], "two_sided")
// Error: `alternative` must be either 'two_side', 'greater', or 'less'.
permutationTest([1, 2, 3], [4, 5, 6], "two_tail")
// Error: `alternative` must be either 'two_side', 'greater', or 'less'.
```

Both documented names are corrected to `two_side`, the value the code branches on.

## What is deliberately unchanged

The accepted string itself. `'two_side'` has been the public API since the function landed, so renaming the code to match the docs would break existing callers to gain a nicer spelling. The p-value math is also untouched — the numerical question in #469 was settled in the thread (the library's two-sided value is correct; the discrepancy belonged to the Python package).

## Testing

`pnpm test` on the branch, re-run 2026-08-15: tsc, biome, documentation lint, and the node:test suite all pass — **294/294 tests, 0 failures**. The branch is based on `8449cf5`; it merges cleanly into current `main` (`0c33928`), verified with `git merge-tree --write-tree`.

Documentation-only, so no failing-before test is possible in this repo; the fail-before evidence is the repro above, run against code identical to `main` at that base.

---

### DIFF
--- a/src/permutation_test.js
+++ b/src/permutation_test.js
@@ -13,7 +13,7 @@ import shuffleInPlace from "./shuffle_in_place.js";
  *
  * @param {Array<number>} sampleX first dataset (e.g. treatment data)
  * @param {Array<number>} sampleY second dataset (e.g. control data)
- * @param {string} alternative alternative hypothesis, either 'two_side' (default), 'greater', or 'less'
+ * @param {'two_side'|'greater'|'less'} [alternative='two_side'] alternative hypothesis
  * @param {number} k number of values in permutation distribution.
  * @param {Function} [randomSource=Math.random] an optional entropy source
  * @returns {number} p-value The probability of observing the difference between groups (as or more extreme than what we did), assuming the null hypothesis.
--- a/src/permutation_test.js
+++ b/src/permutation_test.js
@@ -6,14 +6,14 @@ import shuffleInPlace from "./shuffle_in_place.js";
  * to determine if two data sets are *significantly* different from each other, using
  * the difference of means between the groups as the test statistic.
  * The function allows for the following hypotheses:
- * - two_tail = Null hypothesis: the two distributions are equal.
+ * - two_side = Null hypothesis: the two distributions are equal.
  * - greater = Null hypothesis: observations from sampleX tend to be smaller than those from sampleY.
  * - less = Null hypothesis: observations from sampleX tend to be greater than those from sampleY.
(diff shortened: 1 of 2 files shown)


src/permutation_test.js
```diff
@@ -6,14 +6,14 @@ import shuffleInPlace from "./shuffle_in_place.js";
...
- * - two_tail = Null hypothesis: the two distributions are equal.
+ * - two_side = Null hypothesis: the two distributions are equal.
  * - greater = Null hypothesis: observations from sampleX tend to be smaller than those from sampleY.
  * - less = Null hypothesis: observations from sampleX tend to be greater than those from sampleY.
  * [Learn more about one-tail vs two-tail tests.](https://en.wikipedia.org/wiki/One-_and_two-tailed_tests)
  *
  * @param {Array<number>} sampleX first dataset (e.g. treatment data)
  * @param {Array<number>} sampleY second dataset (e.g. control data)
- * @param {string} alternative alternative hypothesis, either 'two_sided' (default), 'greater', or 'less'
+ * @param {string} alternative alternative hypothesis, either 'two_side' (default), 'greater', or 'less'
```
tmcw: Hmm, this should probably be an enum type instead of a string?

MohammedAlkindi: Done in ea267aa: `@param {'two_side'|'greater'|'less'} [alternative='two_side']`. `documentation build` now emits a UnionType of three string literals and picks up the default, where it previously emitted a bare `string`.

The `.d.ts` has a worse version of the same problem. `src/permutation_test.d.ts` names the third parameter `string` and types it `string`, so TypeScript users see `permutationTest(x, y, string?: string, ...)` with no hint of the accepted values:

```ts
alternative?: "two_side" | "greater" | "less"
```

I kept that out of this PR because it narrows a shipped public type, which is a different call from a docs fix.

`tsc --noEmit`, `biome check` and `documentation lint src` clean, 294 tests pass.

tmcw (approved):
Thanks!

PR #829, opened 2026-08-15
feat(probability): add normalDistribution (NORM.DIST equivalent)

### LINKED ISSUE #544: What would the equivalent of NORM.DIST be?
https://support.microsoft.com/en-us/office/norm-dist-function-edb1cc14-a21c-4e53-839d-8082074c9f8d

```
NORM.DIST(x,mean,standard_dev,cumulative)

Returns the normal distribution for the specified mean and standard deviation. This function has a very wide range of applications in statistics, including hypothesis testing.

The NORM.DIST function syntax has the following arguments:

X     Required. The value for which you want the distribution.

Mean     Required. The arithmetic mean of the distribution.

Standard_dev     Required. The standard deviation of the distribution.

Cumulative     Required. A logical value that determines the form of the function. If cumulative is TRUE, NORM.DIST returns the cumulative distribution function; if FALSE, it returns the probability density function.
```

MohammedAlkindi:
Closes https://github.com/simple-statistics/simple-statistics/issues/544

## Summary

The library has no normal probability density function, and the cumulative route most users find first (`cumulativeStdNormalProbability`) is a four-decimal table lookup with z rounded to the nearest 0.01 — off by up to 2.0e-3 against numerically integrated ground truth (measured in #544). This adds the NORM.DIST equivalent invited there in 2022.

`normalDistribution(x, mean, standardDeviation, cumulative)` mirrors the spreadsheet signature: the probability density by default, the cumulative probability via `0.5 * (1 + errorFunction(z / Math.SQRT2))` when `cumulative` is true — within 4.2e-8 of ground truth over z ∈ [−4, 4], versus 2.0e-3 for the table route. Excel's documented examples check out: `normalDistribution(42, 40, 1.5, true)` → 0.90878878 (Excel documents 0.9087888) and `...false` → 0.10934005 (Excel documents 0.10934).

## Design decisions

- `cumulative` is a default parameter (`cumulative = false`), and input types are not validated, per review. An earlier revision threw on a non-boolean `cumulative`; that check and its test were removed in aca2165.
- `standardDeviation <= 0` throws, in the repo's `FUNCTION requires …` error style.
- Deliberately unchanged: `cumulativeStdNormalProbability`, `standardNormalTable`, and `errorFunction`. This adds the missing function; it does not alter existing routes.

## Limitations

Cumulative precision is inherited from `errorFunction`'s Abramowitz–Stegun-type approximation (absolute error below 1.5e-7). In particular `normalDistribution(mean, mean, sd, true)` returns 0.499999985 rather than exactly 0.5, because the approximation gives erf(0) ≈ 3.1e-8. The test at the mean point asserts within 1e-7 and documents this rather than papering over it.

## Testing

Fail-before: the function does not exist on `main` — `git show upstream/main:index.js` contains no reference to `normal_distribution`.

Pass-after, re-run at `aca2165`: `npm test` runs tsc, biome, documentation lint, and the node:test suite — **302/302 tests (294 existing + 8 new), 0 failures**. Reference values were computed independently of the library's own `errorFunction`, using the C standard library's `erf` (via Python's `math.erf`), plus Excel's documented NORM.DIST examples.

Merges cleanly into current `main` (`7643ea9`); GitHub reports MERGEABLE/CLEAN and all three checks pass at `aca2165`.

---

### DIFF
--- a/src/normal_distribution.js
+++ b/src/normal_distribution.js
@@ -1,0 +1,55 @@
+import errorFunction from "./error_function.js";
+
+/**
+ * **[Normal distribution](https://en.wikipedia.org/wiki/Normal_distribution)**
+ *
+ * Returns either the probability density or the cumulative probability of a
+ * normal distribution with the given mean and standard deviation, evaluated
+ * at x. It is the equivalent of the NORM.DIST function found in spreadsheet
+ * software.
+ *
+ * The cumulative form is computed with `errorFunction`, so it is more precise
+ * than converting x to a z-score and looking it up in the four-decimal
+ * `standardNormalTable` used by `cumulativeStdNormalProbability`.
+ *
+ * @param {number} x the value for which you want the distribution
+ * @param {number} mean the mean of the distribution
+ * @param {number} standardDeviation the standard deviation of the distribution
+ * @param {boolean} [cumulative=false] if true, return the cumulative
+ * distribution function; if false, return the probability density function
+ * @returns {number} the probability density at x, or the probability of a
+ * value at most x
+ * @throws {Error} if standardDeviation is not greater than zero
+ * @throws {Error} if cumulative is neither true nor false
+ * @example
+ * normalDistribution(42, 40, 1.5, true).toFixed(4); // => '0.9088'
+ * normalDistribution(42, 40, 1.5, false).toFixed(4); // => '0.1093'
+ */
+function normalDistribution(x, mean, standardDeviation, cumulative) {
+    // Set default arguments
+    if (cumulative === undefined) {
+        cumulative = false;
+    }
+    if (cumulative !== true && cumulative !== false) {
+        throw new Error(
+            "normalDistribution requires cumulative to be true or false"
+        );
+    }
+    if (standardDeviation <= 0) {
+        throw new Error(
+            "normalDistribution requires a standard deviation greater than zero"
+        );
+    }
+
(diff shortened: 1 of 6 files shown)


src/normal_distribution.js
```diff
@@ -0,0 +1,55 @@
...
+function normalDistribution(x, mean, standardDeviation, cumulative) {
+    // Set default arguments
+    if (cumulative === undefined) {
+        cumulative = false;
+    }
+    if (cumulative !== true && cumulative !== false) {
+        throw new Error(
+            "normalDistribution requires cumulative to be true or false"
+        );
+    }
```
tmcw: This should use a JavaScript default parameter (`cumulative = false`) instead of doing it by hand in the method. And we should not validate the types of inputs because they are documented and specified in types.

tmcw (changes requested):
Looks good, just a minor parameter type / default change.

MohammedAlkindi:
Done, both. `cumulative = false` is a default parameter now, and the boolean check is gone along with its test and the `@throws` line.

One thing beyond the literal ask, in case you'd rather I didn't: I also changed `cumulative === true` to `if (cumulative)`. With the check removed, the strict comparison would quietly return the density for a truthy non-boolean, which is the mistake the check used to catch.

tmcw (approved):

tmcw:
Thanks!

PR #832, opened 2026-08-15
fix(types): declare jenks return type as number[] | null

MohammedAlkindi:
## Summary

`jenks(data, nClasses)` returns `null` whenever `nClasses` is greater than the number of data points — since the function's introduction — but the type declaration promises `number[]`. Under `strictNullChecks` a consumer dereferences the result with no warning and hits a runtime `TypeError`. Same class as #823. The declaration is now `number[] | null`, and the docstring states the null case.

## What is deliberately unchanged

Runtime behavior, and the null contract itself — whether returning null rather than throwing is the right design is a separate question this PR does not touch.

## Testing / limitations

The repo's `tsconfig` has no `strictNullChecks`, so **no failing-before test is possible in-repo** (`test/types.ts` cannot discriminate this). A discriminating consumer compiled with `tsc --strict` shows it instead: before, an unguarded `jenks([1, 2, 3], 5).length` compiles silently — and that call returns `null` at runtime (verified) — after, it fails with `TS2531: Object is possibly 'null'`, and a null-guarded version compiles.

`pnpm test` on the branch, re-run 2026-08-15: **301/301 tests, 0 failures**. Based on current `main` (`0c33928`). Patch changeset included.

---

### DIFF
--- a/.changeset/jenks-null-return-type.md
+++ b/.changeset/jenks-null-return-type.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Declare that `jenks` can return `null`. The implementation returns `null` whenever `nClasses` is greater than the number of data points — since the function's introduction — but the type declaration promised a plain `number[]`, so TypeScript consumers under `strictNullChecks` got no warning before dereferencing the result. Runtime behavior is unchanged; only the declaration and docstring now state the `null` case.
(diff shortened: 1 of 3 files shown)


tmcw (approved):
Probably do want to throw instead of null here, but it should wait for a major version bump.

PR #837, opened 2026-08-20
fix(geometric-mean): keep the mean finite when the product leaves range

MohammedAlkindi:
## What broke

`geometricMean` multiplies every element into a running product and only takes the n-th root at the end, so the intermediate product leaves double range long before the answer does. The function returns `Infinity` or `0` for inputs whose geometric mean is perfectly representable.

The clearest way to see it is that this repo already contains the correct algorithm. `logAverage` in `src/log_average.js` computes the same mathematical quantity in log space and does not overflow:

| input | `geometricMean` | `logAverage` |
|---|---|---|
| `[4, 9]` | 6 | 6 |
| `[1,2,3,4,5]` | 2.605171084697352 | 2.6051710846973517 |
| 500 x `1e10` | **Infinity** | 9999999999.998085 |
| 200 x `1e300` | **Infinity** | 9.999999999980437e+299 |
| 500 x `1e-10` | **0** | 1.0000000000001915e-10 |

They agree on every in-range input and diverge exactly where the naive product leaves range. The true geometric mean of 500 copies of `1e10` is `1e10`.

## The change

The arithmetic is not replaced. The original product path stays, and a log-space fallback runs only when the intermediate product actually leaves range, meaning it hit `0` or a non-finite value. Every input whose product is representable keeps its **bit-identical** previous result.

That matters here because the existing tests assert exact equality, for example `assert.equal(ss.geometricMean([4, 1, 1/32]), 0.5)`. A straight log-space rewrite returns `0.4999999999999999` and breaks the suite. Guarding the fallback means there is no precision change to disclose at all, rather than a 1-ulp shift to apologise for.

Zero is distinguished from underflow explicitly, so `geometricMean([0, 1, 2])` still returns 0.

## Verified

Node v24.18.0, Windows 11.

- The new test fails on unmodified `src/geometric_mean.js` with `expected about 1e10, got Infinity`, and passes with the change. I checked that by restoring only that file from `main` and re-running, so the failure is real and not a stale import.
- Full suite on this branch: **302 tests, 302 pass, 0 fail**.
- `npm run lint` clean, including the `documentation lint` postlint step.
- Differential run against the unpatched original over **200,000 generated inputs**, lengths 1 to 8, magnitudes spanning `1e-300` to `1e300` plus exact `0` and `1`: of the 108,748 inputs where the original returned a usable finite non-zero value, the patched version returned an `Object.is` identical result for **all 108,748**. 40,102 inputs that previously returned `Infinity` or `0` now return a finite value. **Zero regressions.**
- A changeset is included, `patch`.

Values were checked against exact BigInt arithmetic with an integer Newton square root, so the reference could not share the bug under test, plus closed-form cases that need no reference at all: the geometric mean of n copies of v is exactly v.

### DIFF
--- a/.changeset/geometric-mean-product-range.md
+++ b/.changeset/geometric-mean-product-range.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Fix `geometricMean` returning `Infinity` or `0` for inputs whose geometric mean is perfectly representable. The running product left floating-point range long before the mean did, so 500 values of `1e10` overflowed the product to `Infinity` and returned `Infinity` instead of `1e10`, and 500 values of `1e-10` underflowed it to `0` and returned `0` instead of `1e-10`. The product is now only used when it stays in range; when it overflows or underflows the mean is computed in log space instead, matching the approach `logAverage` already uses for the same quantity. Inputs whose product is representable keep their exact previous result, and an input containing a zero still returns `0`.
--- a/src/geometric_mean.js
+++ b/src/geometric_mean.js
@@ -39,6 +39,7 @@ function geometricMean(x) {
 
     // the starting value.
     let value = 1;
+    let containsZero = false;
 
     for (let i = 0; i < x.length; i++) {
         // the geometric mean is only valid for positive numbers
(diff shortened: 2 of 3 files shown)


tmcw (commented):
I'm not sure that this one makes that much sense: the logAverage method is there and documented as the preferred option for large and small products. For geometricMean to switch its implementation based on the size of the inputs feels too "magic" for a library that should be quite literal in terms of how it does the math. This also makes the complexity of geometricMean more complicated because it includes this case where certain inputs cause it to compute the log twice and to use a different algorithm.

I think that users should opt for the log average when using large inputs, as we document, instead of the implementation silently changing based on the magnitude of the inputs.

PR #841, opened 2026-08-24
test: add factorial edge case test for 171!

OlgaEle:


OlgaEle:
Hi @tmcw, thank you for your reviews!

Maybe we can just keep only this for now? And close the PR #843

tmcw (approved):

PR #843, opened 2026-08-24
Feature/linear regression pearson r

OlgaEle:


test/linear_regression.test.js
```diff
@@ -58,7 +58,19 @@ describe("linear regression", function () {
...
+        const result = linearRegression([
+            [1, 2],
+            [2, 4],
+            [3, 5],
+            [4, 4],
+            [5, 5]
+        ]);
+        console.log("Actual r value:", result.r);
+        assert.ok(true);
+    });
```
tmcw: This test doesn't assert anything specific - it should probably be removed.

test/factorial.test.js
```diff
@@ -22,4 +22,7 @@ describe("factorial", function () {
     it("can calculate 100! = 1", function () {
         assert.equal(ss.factorial(100), 9.33262154439441e157);
     });
+    it("returns Infinity for numbers too large to represent safely", function () {
+        assert.equal(ss.factorial(171), Infinity);
+    });
```
tmcw, line 27: This is a fine addition to the tests but isn't related to the rest of the PR

tmcw (changes requested):

test/factorial.test.js
OlgaEle: Added a comment [here](https://github.com/simple-statistics/simple-statistics/pull/841)

MohammedAlkindi:
This conflicts because of #857, which replaced the raw-sum accumulation in `linearRegression` with one centered on the means. `sumXX` and `sumXY` are computed from deviations now, so the `sumYY` here has nothing to rebase onto. Sorry, that was mine.

The same `r` falls out of the loop that is there now, without another pass: name the y deviation, accumulate `sumYY += deviationY * deviationY` beside `sumXX`, and `r` is `sumXY / Math.sqrt(sumXX * sumYY)`. I checked that against `sampleCorrelation` on four datasets including one offset to 1e8 and it agrees to 1e-12.

That form also avoids the cancellation #857 was fixing, which a raw `sumYY` would have brought back.

PR #844, opened 2026-08-29
docs(jsdoc): make kMeansCluster's example reproducible

MohammedAlkindi:
## Summary

`kMeansCluster`'s `@example` calls the function without a `randomSource`, so it falls back to `Math.random`. The initial centroid pick then decides which point seeds cluster 0 vs cluster 1, so running the example exactly as written matches the documented `{labels: [0, 1], ...}` output only about half the time; the other half it comes back reversed.

## Changes

- Pass a fixed `randomSource` in the example, the same technique the test suite already uses for this exact input (`test/k_means_cluster.test.js`'s `nonRNG`).
- Add a test pinning that literal example call and its output.
- Patch changeset. No change to `kMeansCluster` itself.

## Testing

Ran the original example 200 times: 88/200 did not match the documented output. With the fixed `randomSource`, 200/200 match. `pnpm test` (tsc, biome, `documentation lint`, node:test): 311/311 pass, up from 310.

### DIFF
--- a/.changeset/kmeanscluster-example-reproducible.md
+++ b/.changeset/kmeanscluster-example-reproducible.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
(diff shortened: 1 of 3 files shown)


tmcw (commented):
I think that this would be a confusing example - most people _will_ want the randomized behavior, and adding a fixed random source would lead to people copying & pasting an atypical example. Right now, examples are not unit-tested - though they could be, and if we get to that point, this would make more sense. But for now, I don't think it's a good tradeoff.

PR #849, opened 2026-08-31
fix(silhouette): exclude the point itself from its own cluster mean

MohammedAlkindi:
`a(i)` sums over every member of the point's own cluster, including the point itself, then divides by the full cluster size:

```js
for (let i = 0; i < group.length; i++) total += distances[which][group[i]];
return total / group.length;
```

The self-distance is zero, so `a(i)` is low by `(n-1)/n` and every silhouette value in a cluster larger than one point is inflated. On the fixture already in the suite, `[[0.2],[0.4],[0.6],[0.8]]` labelled `[0,0,1,1]`:

```
before    [0.8,  0.667, 0.667, 0.8]
after     [0.6,  0.333, 0.333, 0.6]
sklearn   [0.6,  0.333, 0.333, 0.6]
```

The same helper computes `b(i)`, where the point is never in the group, so the exclusion is opt-in via a fourth argument and the `b(i)` call site is unchanged.

The existing test asserted the old values, and its comment reads `Outer points have a = 0.1, b = 0.5`. That 0.1 is the understated figure, so the expectation came from the implementation. New values are derived from the definition and cross-checked against `sklearn.metrics.silhouette_samples`.

`npm test`: 313 passing before, 314 after, nothing else changed status.

Not addressed here: `NaN` when identical points sit in different clusters, and a throw on a label with no members. Both predate this.

### DIFF
--- a/src/silhouette.js
+++ b/src/silhouette.js
@@ -27,7 +27,8 @@ function silhouette(points, labels) {
             const a = meanDistanceFromPointToGroup(
                 i,
                 groupings[labels[i]],
-                distances
+                distances,
+                true
             );
             const b = meanDistanceToNearestGroup(
                 i,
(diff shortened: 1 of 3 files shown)


tmcw (changes requested):
At least according to what I'm cross-referencing [on Wikipedia](https://en.wikipedia.org/wiki/Silhouette_(clustering)#Definition), the silhouette should _never_ include a comparison from the point to itself, so maybe this is better off as a bugfix rather than a new option?

MohammedAlkindi:
Agreed, and it makes the change smaller. Dropped the option in 9348a16; the point is now left out unconditionally.

The whole diff is five lines inside `meanDistanceFromPointToGroup`. No signature change and no call site change, because the same helper computes both terms and the point is never a member of another cluster, so nothing is skipped on the b(i) path. `silhouetteMetric([[0.25], [0.75]], [0, 0])` is still 1.

Values are what they were: `[0.6, 1/3, 1/3, 0.6]` on the fixture already in the suite, and `[0, 0.9, 10/11]` for a singleton cluster, both matching `sklearn.metrics.silhouette_samples`.

I had made it opt-in only to keep the b(i) call provably untouched. Reading it against the definition the way you did, that caution was buying nothing.

One thing I did not add back: the singleton guard. With the point excluded, a cluster of one has nothing to average, but `silhouette` already assigns those points 0 before calling this, and no other caller exists. Happy to put an explicit guard in if you would rather it not depend on that.

`npm test`: 314 passing, 0 failing.

src/silhouette.js
```diff
@@ -117,8 +117,13 @@ function meanDistanceToNearestGroup(which, labels, groupings, distances) {
...
 /**
- * Calculate the mean distance between a point and all the points in a group
- * (possibly its own).
+ * Calculate the mean distance between a point and the other points in a
+ * group, which may be its own.
+ *
+ * A point is never its own neighbour, so it takes no part in the sum or the
+ * count. When `group` is the point's own cluster that makes this a(i); when
+ * `group` is any other cluster the point is not a member of it and nothing
+ * is skipped, which makes this b(i).
```
tmcw: I think we should drop this comment: it's explaining why this isn't a behavior - this is kind of an LLM writing tic, fwiw, so good to keep an eye on assuming that you're generating a lot of this stuff.

test/silhouette.test.js
```diff
@@ -36,15 +36,48 @@ describe("silhouette test", function () {
...
+    // every member of the nearest other cluster.
+    //
+    // Outer points (0.2 and 0.8) each have one other member in their own
+    // cluster: a = 0.2. Their nearest-cluster mean is b = (0.4 + 0.6) / 2 = 0.5.
+    // s = (b - a) / max(a, b) = (0.5 - 0.2) / 0.5 = 0.6.
+    //
+    // Inner points (0.4 and 0.6) each have one other member in their own
+    // cluster: a = 0.2. Their nearest-cluster mean is b = (0.2 + 0.4) / 2 = 0.3.
+    // s = (0.3 - 0.2) / 0.3 = 1/3.
+    //
```
tmcw:
```suggestion
```

This is kind of excessive commenting (an LLM thing, I'm guessing)

tmcw (changes requested):

src/silhouette.js
```diff
@@ -117,8 +117,8 @@ function meanDistanceToNearestGroup(which, labels, groupings, distances) {
 }
 
 /**
- * Calculate the mean distance between a point and all the points in a group
- * (possibly its own).
+ * Calculate the mean distance between a point and the other points in a
+ * group, which may be its own.
```
tmcw: It's not clear what this is saying now - aren't we _not_ comparing a point to itself?

tmcw (approved):

PR #850, opened 2026-08-31
fix(inverse-error-function): refine the estimate with Newton's method

MohammedAlkindi:
`inverseErrorFunction` returns the Winitzki approximation and stops there. Its round-trip error through `errorFunction` reaches 3.3e-4, and the existing test in `error_function.test.js` allows 4e-4, so it passes with about 18% of the tolerance to spare.

`probit` is built on it and inherits the error at full size:

```
probit(0.975)   1.9571619317   published 1.9599639845
probit(0.99)    2.3213677049   published 2.3263478740
probit(0.999)   3.0801321223   published 3.0902323062
```

The 95% two-sided z is the value most users of `probit` reach for, and it comes back as 1.957.

`errorFunction` itself is accurate to 7.9e-8 here, so the seed can simply be refined against it. Two Newton steps, using `2 / sqrt(pi) * exp(-x^2)` for the derivative:

```
round trip max     3.284e-4  ->  8.301e-8
probit max error   1.010e-2  ->  5.669e-8
```

`probit(0.975)` is now 1.9599639471.

Infinity at the ends, `NaN` outside the domain, oddness, and `inverseErrorFunction(0) === 0` are all preserved. The zero case is guarded, because `errorFunction(0)` is off by 3e-8 and the refinement would otherwise walk an exact answer away from zero.

No existing expectation changed. `npm test` goes 313 passing to 318, 0 failing.

Reference values in the new test come from inverting a Maclaurin series for `erf` by bisection, residual ~1e-16.

Not addressed: the `21 * epsilon` tolerance in `cumulative.test.js`. That slack is the 0.01 granularity of `standardNormalTable`, not this function. It measures 2.000e-3 both before and after.

### DIFF
--- a/src/inverse_error_function.js
+++ b/src/inverse_error_function.js
@@ -16,24 +16,23 @@ const TWO_OVER_SQRT_PI = 2 / Math.sqrt(Math.PI);
  * @returns {number} estimated inverted value
  */
 function inverseErrorFunction(x) {
+    const a = (8 * (Math.PI - 3)) / (3 * Math.PI * (4 - Math.PI));
+
+    const inv = Math.sqrt(
+        Math.sqrt(
+            Math.pow(2 / (Math.PI * a) + Math.log(1 - x * x) / 2, 2) -
+                Math.log(1 - x * x) / a
+        ) -
+            (2 / (Math.PI * a) + Math.log(1 - x * x) / 2)
+    );
+
     // `errorFunction` is off by a few parts in 1e8 at zero, so the
     // refinement below would walk the exact answer away from it.
     if (x === 0) {
         return 0;
     }
 
-    const a = (8 * (Math.PI - 3)) / (3 * Math.PI * (4 - Math.PI));
-
-    const logTerm = Math.log(1 - x * x);
-    const halfLog = 2 / (Math.PI * a) + logTerm / 2;
-
-    let estimate = Math.sqrt(
-        Math.sqrt(halfLog * halfLog - logTerm / a) - halfLog
-    );
-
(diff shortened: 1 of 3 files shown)


src/inverse_error_function.js
```diff
@@ -1,27 +1,62 @@
...
-            Math.pow(2 / (Math.PI * a) + Math.log(1 - x * x) / 2, 2) -
-                Math.log(1 - x * x) / a
-        ) -
-            (2 / (Math.PI * a) + Math.log(1 - x * x) / 2)
+    const logTerm = Math.log(1 - x * x);
+    const halfLog = 2 / (Math.PI * a) + logTerm / 2;
+
+    let estimate = Math.sqrt(
+        Math.sqrt(halfLog * halfLog - logTerm / a) - halfLog
     );
```
tmcw: This is equivalent math to `inv`. Maybe at some point it'd be useful to save the math ops, but I think combined with a functional change we should keep the computation of `inv`/`estimate` the same to minimize diff and make it clear what's changing.

tmcw (changes requested):

MohammedAlkindi:
Done. The expression for `inv` is now byte for byte what it was, and the Newton steps run on a copy, so the diff is only the refinement plus the import.

Values are unchanged: `probit` still lands within 5.669e-8 of the published quantiles, `inverseErrorFunction(0)` is still exactly 0, the ends still give ±Infinity, out of domain still gives NaN, and oddness still holds exactly.

`npm test`: 318 passing, 0 failing.

PR #851, opened 2026-08-31
fix(chi-squared): correct the critical value at 7 degrees of freedom

MohammedAlkindi:
`chiSquaredDistributionTable[7][0.99]` carries 1.25. The exact value is 1.23904, which rounds to 1.24 at the table's two decimal places. Put the other way round, 1.25 sits at an upper tail of 0.98973, not 0.99.

`chiSquaredGoodnessOfFit` compares its statistic against this cell, so a statistic in [1.24, 1.25) came back as not significant when the correct answer is the opposite. Neighbouring rows already agree: 0.87209 and 1.64650 at 6 and 8 degrees of freedom round to the tabulated 0.87 and 1.65.

Verified two independent ways, bisection on the regularized incomplete gamma and Simpson integration of the density. The solver reproduces 3.8415, 18.3070 and 37.6525 for the 0.05 column at 1, 10 and 25 degrees of freedom.

A correction to #840: its changeset called df=25 "the only cell out of step". That was too strong, and this entry disproves it. I have removed the same claim from the test comment.

Nine further cells disagree at exact two-decimal rounding, all by 0.0050 to 0.0079:

```
df=6   p=0.1    10.65  exact 10.64464
df=8   p=0.005  21.96  exact 21.95495
df=15  p=0.975   6.27  exact  6.26214
df=18  p=0.9    10.87  exact 10.86494
df=27  p=0.005  49.65  exact 49.64492
df=28  p=0.99   13.57  exact 13.56471
df=70  p=0.01  100.42  exact 100.42518
df=70  p=0.005 104.22  exact 104.21490
df=90  p=0.05  113.14  exact 113.14527
```

I left those alone. Each is within a half unit in the last place, so they are consistent with Hines and Montgomery rounding from a lower-precision intermediate, and the file cites that book. Only df=7 falls outside that band, at 0.011. Happy to send the nine separately if you would rather the table match exact quantiles than the printed source.

`npm test`: 313 passing to 314, 0 failing.


test/chi_squared_distribution_table.test.js
```diff
@@ -8,7 +8,14 @@ describe("chiSquaredDistributionTable", function () {
...
+        // 35.56 at 26.
         assert.equal(ss.chiSquaredDistributionTable[25][0.1], 34.38);
     });
+
+    it("has the published upper-tail critical value at 7 degrees of freedom", function () {
+        // The exact value with an upper tail of 0.99 is 1.23904, which rounds
+        // to 1.24. The neighbouring rows already agree, 0.87 at 6 degrees of
+        // freedom and 1.65 at 8.
+        assert.equal(ss.chiSquaredDistributionTable[7][0.99], 1.24);
+    });
```
tmcw: I think this is taking TDD a bit too far: asserting that a constant in an array is a value isn't a very useful test, given that it does not have the ability to protect against refactoring regressions and doesn't assert any logic. I think we should probably remove this test case.

tmcw (changes requested):

MohammedAlkindi:
Removed in 5d58cdd. Agreed, it asserts a table constant rather than the output of anything.

The file still carries a one-line change: the comment on the 25 degrees of freedom case claimed that entry "was the only one out of step", which is what I wrote in #840 and what this PR disproves, so I corrected the sentence rather than leave it standing.

By the same reasoning the assertion above is also just a constant check. I left it because it arrived with #840 and is outside this change, but I am happy to drop it and the file with it if you would rather.

`npm test` is back to 313 passing, 0 failing.

tmcw (approved):

PR #852, opened 2026-08-31
fix(gamma): lift small arguments before the Nemes expansion

MohammedAlkindi:
`gamma(1.01)` returns 2.80027. The true value is 0.99433.

Nemes' approximation is asymptotic. Its correction terms are divided by powers of `n - 3/4`, so they stop shrinking once the argument is small, and the series is worthless there:

```
gamma(1.01)   2.80027  exact 0.99433
gamma(1.05)   1.68117  exact 0.97350
gamma(1.10)   1.21285  exact 0.95135
gamma(1.50)   0.88796  exact 0.88623
gamma(1.90)   0.96186  exact 0.96177
```

The error decays smoothly with the argument, so nothing looked obviously broken further out. Negative arguments inherit it through Euler's reflection formula, worst 0.64 relative near -0.01.

Arguments in `(0, 1)` were already lifted by a two-step recurrence, added to escape infinite recursion in the reflection branch. The same lift fixes this; it just needed to run generally rather than only below 1. `gamma(n) = gamma(n + 1) / n`, applied until the argument reaches 20.

```
worst relative error, (0, 30]    1.816    ->  1.9e-13
worst relative error, [-11, 0)   0.645    ->  2.8e-13
```

Reference values come from an independent Lanczos evaluation, checked against `sqrt(pi)` at 0.5 and `-2 * sqrt(pi)` at -0.5 to 1e-15.

Threshold 20 is where the returns flatten: 4 gives 4.5e-8, 12 gives 4.6e-12, 30 gives 4.9e-14 for twelve more divisions.

Integer arguments still short-circuit to `factorial`, and zero and negative integers still return `NaN`.

Two things this changes that are worth flagging. The `@example` values for `gamma(11.5)` and `gamma(-11.5)` recorded the old error to full precision, so they are updated. And the test asserting `gamma(0.5) === gamma(2.5) / (0.5 * 1.5)` was written against the two-step lift; it is now the one-step form, which still holds exactly by construction.

The pinned `gamma(11.54)` expectation is untouched and still passes. Worth noting it is 8.0e-5 away from the true value and the assertion allows 1e-4, so it passes from the other side now.

`npm test`: 313 passing to 315, 0 failing.

### DIFF
--- a/.changeset/gamma-small-argument.md
+++ b/.changeset/gamma-small-argument.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+Correct `gamma` for small arguments. Nemes' approximation is asymptotic, dividing its terms by powers of `n - 3/4`, so the series stops converging once the argument is small. Applied directly it returned 2.8 for `gamma(1.01)` against a true value of 0.994, and 1.2128 for `gamma(1.1)` against 0.951. Arguments in `(0, 1)` were already lifted by a two-step recurrence; that lift now runs generally, raising the argument until the expansion converges. Worst relative error over `(0, 30]` goes from 1.8 to 1.9e-13, and over `[-11, 0)` from 0.64 to 2.8e-13, since negative arguments reach the expansion through Euler's reflection formula. The two `@example` values for `gamma(11.5)` and `gamma(-11.5)` recorded the old error and have been updated.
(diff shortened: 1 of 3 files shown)


tmcw (approved):

tmcw:
Thanks!

PR #853, opened 2026-08-31
feat(skewness): add an opt-in biased coefficient

### LINKED ISSUE #167: Implement biased sample skewness
R and Python both support biased sample skewness - Python's scipy has a Bias=false opt-out, whereas R doesn't have an opt-out. We should support this as well, but as an opt-in in order to avoid a major version difference and follow the existing pattern of making features opt-in.

MohammedAlkindi:
Closes #167.

`sampleSkewness` takes a `biased` flag. `true` returns the population moment ratio `m3 / m2^(3/2)`, which is what R reports and what `scipy.stats.skew` returns by default. The current adjusted Fisher-Pearson coefficient stays the default, so this is opt-in as the issue asks.

```js
sampleSkewness([2, 4, 6, 3, 1]);        // 0.590128656384365
sampleSkewness([2, 4, 6, 3, 1], true);  // 0.3958703373438167
```

The signature follows `normalDistribution(x, mean, standardDeviation, cumulative)`, the existing trailing-flag pattern in the library.

Checked against `scipy.stats.skew` 1.18.0 on six samples, both modes. Worst disagreement is 6.7e-16 for the default and 2.2e-16 for the new path. That run also confirms the current default is exactly scipy's `bias=False`, which the docstring now says.

The two coefficients differ by `sqrt(n * (n - 1)) / (n - 2)`, and there is a test asserting that identity so the branches cannot drift apart.

The default code path is untouched. I ran the six samples before and after the change and the returned doubles are bit-identical, so nothing depending on current behaviour moves.

One thing I did not change: the `n >= 3` requirement still applies in both modes. The biased coefficient is defined at `n = 2`, so that throw is stricter than it needs to be, but relaxing it is a separate decision about the function's contract and the issue does not ask for it.

`npm test`: 313 passing to 316, 0 failing.

### DIFF
--- a/.changeset/biased-sample-skewness.md
+++ b/.changeset/biased-sample-skewness.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": minor
+---
+
+`sampleSkewness` accepts a `biased` flag. Passing `true` returns the biased coefficient, the population moment ratio `m3 / m2^(3/2)`, which is what R reports and what `scipy.stats.skew` returns by default. The existing adjusted Fisher-Pearson coefficient stays the default and is unchanged, so this is opt-in. Closes #167.
--- a/src/sample_skewness.d.ts
+++ b/src/sample_skewness.d.ts
@@ -1,6 +1,6 @@
 /**
  * https://simple-statistics.github.io/docs/#sampleskewness
  */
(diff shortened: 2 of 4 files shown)


tmcw (approved):
Thanks!

PR #854, opened 2026-08-31
feat(root-mean-square): add scaledRootMeanSquare for extreme magnitudes

MohammedAlkindi:
Taking you up on #838, where you closed my in-place fix with:

> I think exposing a _new_ method that uses scaled values would make sense, but to add a silent fallback behavior that re-does the computation in the existing method seems like it is too 'magic' and too implicit

So this adds `scaledRootMeanSquare` and leaves `rootMeanSquare` completely alone. It is the same split the library already has between `geometricMean` and `logAverage`, whose docstring likewise says it is the form "suitable for large or small products".

```
                          rootMeanSquare   scaledRootMeanSquare
[1e200, 2e200, 3e200]     Infinity         2.1602468994692866e+200
[1e-200, 2e-200, 3e-200]  0                2.1602468994692867e-200
[1, 2, 3, 4]              2.7386127875258306   same
[-1, 1, -1, 1]            1                same
```

The exact value for the first row is `sqrt(14/3) * 1e200`, and the returned double is within 1e-15 relative of it. Direct summation breaks above roughly 1e154 and below roughly 1e-162.

Two implementation notes.

`Math.hypot` is scale-safe and would give `Math.hypot(...x) / Math.sqrt(x.length)` in one line, but the spread throws `RangeError: Maximum call stack size exceeded` on a 200,000 element array, so it is not usable for a statistics library. The explicit two-pass loop handles that array fine.

Infinite and `NaN` inputs return what `rootMeanSquare` returns rather than turning into `NaN` through the scaling, and the empty-input throw matches too. There is a test pinning that parity.

The name is the part I am least confident about, so rename it to whatever fits and I will follow.

`npm test`: 313 passing to 318, 0 failing.

### DIFF
--- a/.changeset/scaled-root-mean-square.md
+++ b/.changeset/scaled-root-mean-square.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": minor
+---
+
+Add `scaledRootMeanSquare`, an equivalent way of computing the root mean square for values whose squares leave floating point range. `rootMeanSquare` sums the squares directly, so it returns `Infinity` above roughly 1e154 and `0` below roughly 1e-162. The new function divides through by the largest magnitude first. `rootMeanSquare` is unchanged, following the same split as `geometricMean` and `logAverage`.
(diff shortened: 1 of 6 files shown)


tmcw (approved):
Cool, lgtm - thanks!

PR #855, opened 2026-08-31
feat(regression): add weightedLinearRegression

### LINKED ISSUE #48: Weighted linear regression
Hi

Do you, by any chance, know where I can find a Javascript library for weighted linear regression? If not, would you consider adding it to simple-statistics?

Thanks

Marcus

MohammedAlkindi:
Closes #48, where you said you would accept a patch for this. The references you gave there, R's `lm()` and the weighted least squares article, are what it follows.

`linearRegression` is untouched. This is a separate function, so it does not collide with #843.

Checked against `statsmodels.api.WLS` on six datasets, including unequal weights, a zero weight, fractional weights and coordinates at 1e6. Worst slope disagreement is 1.4e-15 and worst intercept disagreement is 7e-16 relative.

Four identities hold exactly rather than approximately, and each has a test:

```
equal weights            == linearRegression                    exact
weights [0,1,1,1,1]      == linearRegression on the last four   exact
weights [1,2,1]          == linearRegression with point 2 twice exact
one point                => { m: 0, b: y }, matching linearRegression
```

Scaling every weight by the same factor leaves the line unchanged to 1e-12, which is a tolerance rather than an equality because the sums are reassociated.

Validation goes through `validateWeightedInput`, the helper `weightedMean` and the others already use, so the error messages and the empty, length-mismatch, negative-weight and all-zero-weight cases match the rest of the weighted family for free.

Implementation note: it centres on the weighted means before accumulating the cross products rather than summing raw `x*y`, which is why the 1e6 case still agrees to the last couple of bits.

`npm test`: 313 passing to 320, 0 failing.

### DIFF
--- a/.changeset/weighted-linear-regression.md
+++ b/.changeset/weighted-linear-regression.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": minor
+---
+
+Add `weightedLinearRegression`, which fits a line by [weighted least squares](https://en.wikipedia.org/wiki/Weighted_least_squares), the same thing R's `lm()` does when given a `weights` argument. `linearRegression` is the special case where every weight is equal, and it is unchanged. Validation reuses the same helper as the other weighted functions. Closes #48.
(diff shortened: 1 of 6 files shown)


tmcw (approved):
Cool, lgtm, thanks!

PR #857, opened 2026-09-01
fix(regression): center linearRegression on the means before accumulating

MohammedAlkindi:
`linearRegression` accumulates raw sums for the normal equations, so the variance is lost to cancellation once the x values share a large offset. On a perfect line the slope comes back 22% low at x near 1e8, changes sign at 1e9, and is `Infinity` at 1e10. Timestamps hit this in ordinary use: with `Date.now()` as x the slope is 69% wrong.

This accumulates the cross products from deviations about the means instead, which is what `weightedLinearRegression` and `sampleVariance` already do here. It costs a second pass over the data.

`weightedLinearRegression`'s documentation says `linearRegression` is the special case where every weight is equal. The two disagreed at large x; the second added test pins them together.

Before:

```
actual: Infinity,
expected: 2,
```

After: 335 passing, 0 failing, up from 333 on `main` plus the two added tests. `npm run lint` is clean.

### DIFF
--- a/.changeset/linear-regression-cancellation.md
+++ b/.changeset/linear-regression-cancellation.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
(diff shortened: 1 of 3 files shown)


tmcw (approved):
Thanks!

PR #858, opened 2026-09-01
docs(mode): describe how each implementation actually breaks ties

MohammedAlkindi:
`mode`, `modeSorted` and `modeFast` all promise "in the event of a tie, this algorithm will return the most recently seen mode". None of them does, and they do not agree with each other.

`mode` and `modeSorted` sort their input, so the order values arrived in cannot matter and a tie resolves to the smallest tied value. `modeFast` walks the input keeping a strict maximum, so a tie resolves to whichever value reaches the highest count first. `mode([5, 5, 2, 2])` is `2` while `modeFast([5, 5, 2, 2])` is `5`, and the existing tests never catch this because their tied inputs are already in ascending order.

This documents each rule and points at the difference, since the three are presented as interchangeable. No behaviour changes, so there is no failing case to show; the added tests pin all three so the docs cannot drift again. 336 passing, up from 333.

If you would rather the three agreed, that is a behaviour change and I am glad to send it separately.


src/mode_sorted.js
```diff
@@ -1,8 +1,10 @@
 /**
  * The [mode](https://en.wikipedia.org/wiki/Mode_%28statistics%29) is the number
  * that appears in a list the highest number of times.
- * There can be multiple modes in a list: in the event of a tie, this
- * algorithm will return the most recently seen mode.
+ * There can be multiple modes in a list. This algorithm sorts its input, so a
+ * tie resolves to the smallest of the tied values regardless of the order they
+ * appeared in. `modeFast` breaks ties differently, returning whichever tied
```
tmcw: This is a misleading way to describe modeSorted - it expects sorted input, but does not sort its input.

tmcw (changes requested):

tmcw (approved):
Thanks!

PR #859, opened 2026-09-02
fix(quantile-rank): throw on empty input instead of returning NaN

MohammedAlkindi:
`quantileRank([], 3)` and `quantileRankSorted([], 3)` return `NaN`. Every other function here rejects empty input with an error — `quantile`, `min`, `mode`, `rootMeanSquare` — so the `NaN` travels on through downstream arithmetic instead of surfacing at the call site.

Both now throw, matching that convention. `quantileRank` picks the check up through `quantileRankSorted`; the JSDoc records it on both.

With the source change reverted and the two new tests kept:

```
AssertionError [ERR_ASSERTION]: Missing expected exception.
tests 4, pass 2, fail 2
```

With it, 4 passing. Full suite 338 passing, 0 failing, up from 336 on `main` plus the two added tests. `npm test` runs `tsc`, `biome` and `documentation lint` alongside it, all clean.

This does change behaviour for empty input, from a `NaN` return to a throw. I marked the changeset `patch` on the reading that the `NaN` was the anomaly rather than the contract — happy to make it `minor` if you read it the other way.

### DIFF
--- a/.changeset/quantile-rank-empty-input.md
+++ b/.changeset/quantile-rank-empty-input.md
@@ -1,0 +1,5 @@
+---
+"simple-statistics": patch
+---
+
+fix: `quantileRank` and `quantileRankSorted` now throw on empty input instead of returning `NaN`. Every other function in the library rejects empty input with an error, so the `NaN` propagated silently through downstream arithmetic rather than surfacing at the call site.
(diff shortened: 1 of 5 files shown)


tmcw (changes requested):
I think this is better off in a major release - even though the contract is that the arrays are non-empty, throwing on some unexpected input is a pretty big consequence and I don't want to have anyone encounter that and be surprised.

tmcw (approved):
Approving now but I'll merge after we release the next minor version.

PR #860, opened 2026-09-14
fix(jenks): shift by the median before accumulating variance

MohammedAlkindi:
`jenks` puts the same data in different classes depending on its offset. On 12 consecutive millisecond timestamps it splits them 7/3/2; on the same 12 values with the offset removed, 4/4/4. `ckmeans` returns 4/4/4 for both.

`jenksMatrices` accumulates raw sums and computes variance as `sumSquares - (sum * sum) / w`, so the significance is lost to cancellation at that magnitude. Some combinations come back negative, which a sum of squared deviations cannot be: `varianceCombinations[12]` is `[0, 4294967296, 0, -1073741824]`, where the true value is 143. The classing is chosen by comparing those numbers.

This shifts by the median before accumulating and clamps a negative result to zero, which is what `ckmeans` already does. A constant shift does not change a variance, so the cases that were already right are unaffected.

Before, the added test fails:

```
+ actual - expected

+   1700000000007,
+   1700000000010,
-   1700000000004,
-   1700000000008,
```

After: 340 passing, 0 failing, up from 339 on `main`. `npm run lint` is clean.


src/jenks_matrices.js
```diff
@@ -79,6 +82,10 @@ function jenksMatrices(data, nClasses) {
             // of samples.
             variance = sumSquares - (sum * sum) / w;
 
+            if (variance < 0) {
+                variance = 0;
+            }
```
tmcw: I'm not sure what this is fixing and the included test doesn't relate to it - what are we doing here?

MohammedAlkindi:
You're right, and it's removed in 8783b6b.

It was guarding the same cancellation producing a negative intermediate for data with a large spread rather than a large offset, which does happen. But I checked whether it reaches the result and it does not: with and without it, `jenks` returns identical breaks on every case I tried, including the mixed-sign and 1e12-spread ones that generate the negatives. So it fixed nothing observable, had no test, and bundling it with the actual change was wrong.

The PR is now the median shift and the invariance test, nothing else.

tmcw (approved):

PR #865, opened 2026-09-15
fix(weighted-mean): sum through the compensated sum

MohammedAlkindi:
`mean` and `weightedMean` disagree on the same numbers with equal weights. `mean([1e16, 1, -1e16])` is `0.3333333333333333`; `weightedMean` of those values with weights `[1, 1, 1]` returned `0`.

`mean` sums through `sum.js`, which is Kahan-Babuska and says in its own docs that it corrects for floating-point error. `weightedMean` accumulated `x[i] * weights[i]` in a plain loop, so it threw away the correction the library already has. This passes the weighted values through `sum` instead, following the same reuse you asked for on #848.

That costs one temporary array where the loop allocated nothing. I preferred it to a second copy of Kahan-Babuska in this file, and 200k elements measure about 6ms, but say if you would rather have it inlined.

Before, the added test fails:

```
+ actual - expected
+ 0
- 0.3333333333333333
```

After: 340 passing, 0 failing, up from 339. `npm run build` and `npm run lint` are clean.

`weightedVariance` accumulates the same way and is left alone here.


src/weighted_mean.js
```diff
@@ -16,13 +17,8 @@ import validateWeightedInput from "./validate_weighted_input.js";
...
 function weightedMean(x, weights) {
     const totalWeight = validateWeightedInput(x, weights, "weightedMean");
-    let weightedSum = 0;
 
-    for (let i = 0; i < x.length; i++) {
-        weightedSum += x[i] * weights[i];
-    }
-
-    return weightedSum / totalWeight;
+    return sum(x.map((value, i) => value * weights[i])) / totalWeight;
```
tmcw: This would be a performance regression - we don't want to build an equal-length array with `.map` and then throw it away. This can be done with an online algorithm and should

tmcw (changes requested):

MohammedAlkindi:
Done in 31ef588. The correction is applied inside the existing loop now, so there is no intermediate array and no per-call closure.

`weightedMean([1e16, 1, -1e16], [1, 1, 1])` returns `0` on main and `0.3333333333333333` after, matching `mean` on the same input. 340/340 pass under `pnpm test` on Node 24.

I had also missed that CONTRIBUTING.md lists ES5 array methods among the things this project doesn't use, so the `.map` was wrong on that count independently of the allocation.

The tradeoff is a second copy of the Kahan-Babuska correction in this file. I kept it local rather than factoring an accumulator out of `sum.js`, since that touches every caller of `sum`. Happy to do it that way instead if you'd prefer one implementation.

